PR #3, opened 2019-10-23
Nicholas translation

nicholaspung:
Merging in translation files - currently can switch between EN and ES


nicholaspung:
Closing to redo PR. Will be overriding code since there wasn't that many changes.

Issue #59, opened 2019-11-03
Migrate over Donate section


nicholaspung:
### Overview
Migrate content from "Donate" section

### Action Items
Add content from "Donate"

nicholaspung:
Merged

Issue #49, opened 2019-10-30
Migrate content for About page


mmorgs:
Copy content from live site to repo

nicholaspung:
Hey, so I should have created the issue when I started working/finished migrating the content for the About page. My bad that I didn't notice this until today...apologies.

PR #76, opened 2019-11-04
Nicholas/translation

nicholaspung:
Added content for About page and Donate page (no styling) with translation format
Updated the json files housing the translation data.

Will need to figure out how to use i18next when getting data from the backend (if we want admin created/updated FAQs, or if we have Spanish news data)


nicholaspung:
#59, #49 done with migrating content over. Going to say that styling is a different task, if agreed upon!

PR #79, opened 2019-11-06
Map -- initial stakeholders fetch, stateful logic, markers/popup

lucas-homer:
This PR adds some new things to the map. I kept the styling minimal, as those decisions will get ironed out. 
[Demo video here](https://www.loom.com/share/285431c961a846aba0057e406dafeb71)

After this I'd like to get the stakeholders map hooked up to the filters and search params @entrotech built.


lucas-homer:
Another thought -- what kind of info do we want inside the Marker Popup? Do we want to go ahead and throw an `<a>` in there with the website if the stakeholder has one? I guess before I do anything else I can try to flesh out these details.

PR #83, opened 2019-11-10
create useStakeholders hook to set up integrating the list and map views

lucas-homer:
I extracted all the lifecycle/data fetching logic into a hook. This, in my opinion at least, cleans up the stakeholders container component. The search form, stakeholders state, and map, all have their concerns separated now, and the only real state concern in the container component is a little `const [selectedView, toggleSelectedView] = useState(MAP_VIEW)`. Like the StakeholdersList, we can now feed all the stakeholders data down to the Map component. That'll be my next PR after this one. 

* [Demo video to show the search is still working like before](https://www.loom.com/share/b8f5b6d5429247fc92bb513dde4330a7)


lucas-homer:
this needs some more testing... closing this PR for now.

PR #146, opened 2019-11-27
FAQ Service done for both client and server

nicholaspung:
FAQs work with language translation. Added a few extra methods for different use cases.
Currently shows the question/answer that are inputted into the database.


entrotech:
Not sure I understand what your Gatsby proposal is for. Gatsby is a static site generator like Jekyll. We've spent most of our programming time so far moving away from a static site, so we can have a site backed by a real database and allow users to do the data entry and maintenance.

nicholaspung:
My bad for leaving the gatsby-proposal.md file up to be merged. That was meant for me to dump my thoughts for further optimizations later on (once the project was "done", I was going to flesh out the details on adding Gatsby to the project). I'll remove the file from master!

PR #158, opened 2019-12-07
Work on useStakeholder.js hook and StakeholderSearch

entrotech:


client/src/hooks/useStakeholders/useStakeholders.js
```diff
@@ -47,77 +49,94 @@ export function useStakeholders() {
...
+
+    // Exposed to consuming component to execute search
+    search(
+      searchString,
+      selectedLatitude,
+      selectedLongitude,
+      selectedLocationName,
+      selectedCategories,
+      selectedDistance
+    );
```
lucas-homer: this part is interesting... when I pulled down and ran locally, it went ahead and searched based on the initialState lat/long of the LACI coordinates, even though I've permitted the browser to get my location. the `search(...)` is always using the initialState values. this was the tricky part when i initially wrote the custom hook because you want to validate the lat/long, be they initialState (aka LACI) or the verified user location. perhaps it's somewhat moot once we have the geocoder autocomplete. when i did some research on other popular map/search/list-results apps (google maps, yelp, etc), usually they just load up a somewhat 'empty' map without grabbing your location at all, and then you go from there.

lucas-homer: ok I think I see what you're doing here. really just fetching the location so you can have that as a radio button option in the filters, yeah?

client/src/hooks/useStakeholders/useStakeholders.js
```diff
@@ -47,77 +49,94 @@ export function useStakeholders() {
...
-            selectedLongitude,
-            selectedCategories,
-            selectedDistance
-          );
-        }
-      } catch (error) {
-        dispatch({ type: FETCH_FAILURE, error });
-      }
-    };
+  const fetchLocation = async () => {
```
lucas-homer: After looking through what you're doing, I took a stab at simplifying this function a little bit, since we don't really need it to be an async function anymore. We're not waiting on anything really in order to fire off that initial `search(...initialStateParams)` so here's what I ended up with:
```
  const fetchLocation = () => {
    const {
      FETCH_FAILURE,
      FETCH_REQUEST,
      FETCH_SUCCESS,
    } = actionTypes.LOCATION;

    dispatch({ type: FETCH_REQUEST });

    if (navigator.geolocation) {
      navigator.geolocation.getCurrentPosition(
        position => {
          if (!position) {
            dispatch({
              type: FETCH_SUCCESS,
              userCoordinates: { latitude: null, longitude: null },
            });
          }

          const userCoordinates = {
            latitude: position.coords.latitude,
            longitude: position.coords.longitude,
          };
          dispatch({ type: FETCH_SUCCESS, userCoordinates });
        },
        error => {
          dispatch({ type: FETCH_FAILURE, error });
        },
      );
    } else {
      // If browser location permission is denied, the request is
      // "successful", but the result is null coordinates.
      dispatch({
        type: FETCH_SUCCESS,
        userCoordinates: { latitude: null, longitude: null },
      });
    }
  };

```

client/src/components/LocationAutocomplete.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from "react";
+import * as esriService from "../services/esri_service";
+import { TextField, Typography, Button } from "@material-ui/core";
+
+const LocationAutocomplete = props => {
```
lucas-homer: I've found over time I prefer to destructure my props right off the bat, here, so I always have a clear list of props coming in. We should also think about adding prop-types so we can start to solidify things that way too. You can define specific data types like a stakeholder with something like `stakeholder: PropTypes.shape({lat: PropTypes.number.isRequired, ...etc})`

Issue #234, opened 2020-01-08
Create Landing Page


Telos8840:
### Overview
Build out Landing Page based on latest wireframes

### Latest design
https://www.figma.com/file/QgM5vS8NhDNXhknnFpu2Hn/Prototyping-in-Figma?node-id=26%3A394

ExperimentsInHonesty:
@Telos8840 Please provide update:
1. Progress
2. Blocks
3. Availability
4. ETA

ExperimentsInHonesty:
The purpose of this is to copy the current foodoasis.la homepage as closely as possible, while changing out the fake search bar with a real one that takes in the address.

ExperimentsInHonesty:
To be clear, what we are looking for in the mockup is:
1. staying close to the existing design. 
2. Changing the search box to actually be a search box, with the word "Enter an address, neighborhood or zip code"
3. see Tina image in next comment for how current location works
4. Replace the text under the search that currently says 
 - "Food Oasis LA (FOLA) shows you places to find healthy food in Los Angeles – whether you are looking to buy, grow, or need access to free food. 
- "Food Oasis LA (FOLA) has links to food pantries and hot meals" 
5. Leave the "Learn more" link

tina-moh:
![47C405B1-DBAA-46CD-B9EE-AE127AE81B96](https://user-images.githubusercontent.com/56663382/73618602-a2d24680-45dd-11ea-9221-a786dc821880.jpeg)

Telos8840:
This is nearly done. I had some issues initially with Material UI giving me some errors but I just got that resolved today. I'm hoping to have this done by the time we meet next week.

ExperimentsInHonesty:
ETA by next meeting.  Progress going well.

tina-moh:
<img width="675" alt="Landing page 1" src="https://user-images.githubusercontent.com/56663382/74777739-f932bc80-524e-11ea-8d41-7ce85e25dc2a.png">

tina-moh:
<img width="546" alt="Landing Page1" src="https://user-images.githubusercontent.com/56663382/74778347-3b103280-5250-11ea-9f81-079e27565729.png">
<img width="541" alt="Landing page2" src="https://user-images.githubusercontent.com/56663382/74778355-3f3c5000-5250-11ea-8fd3-3927f4831812.png">
  @lucas-homer @Telos8840  please let me know if you have any  comments

tina-moh:
we can have it either in blue or red 
[...]

PR #283, opened 2020-04-07
234 darragh search

entrotech:
Worked mostly on /src/containers/Home/index.js and /components/Search.js to get Home page to call stakeholders with an appropriate url to execute the corresponding query.  I still can't say that Search.js is using the Downshift library correctly. In the process, I stripped out all the persistent global state stuff, since the Home page can just pass the coordinates and place name to the StakeholderContainer component via the url.


entrotech:
Related to issue #234, but may not close it. Handing over to Saul to continue work...

Issue #330, opened 2020-04-28
Results Page Mapbox Component Enhancement


entrotech:
### Overview
Enhance the ResultsMap component to look like the Desktop and Mobile Figma wireframes (linked below).  This is only for the ResultsMap component - the other panels on the page are allocated to different issues. I think the pages labelled Macbook Pro 15, Macbook Pro 16 and Macbook Pro 17 are the wireframes to reference - check with Tina.

### Action Items

- [ ] Study the Figma files below
- [ ] Modify the ResultsMap to mimic the wireframes as closely as possible, including the marker design (the check mark means it's verfied data)
- [ ] There are a few features that aren't shown on the wireframe. For example, clicking on a marker should probably show an pop-up info window near the marker with cursory information about the selected organization. Check with the UX/UI folks about this before proceeding - maybe they don't want this feature, or maybe clicking on a marker highlights the clicked organization on the list in the left-hand panel.
- [ ] Some of the other wireframes show an oversize marker that is the location specified in the serach - I'm not sure if that is still part of the design.
- [ ] The color of the marker should reflect the organization category.

### Resources/Instructions
[Wireframes for Desktop ](https://docs.google.com/document/d/1NaMPWsWJvcTivgS6M6fuJPjdusj8TGMSmw9Pqk3ATbs/edit)

[Wireframe for Mobile](https://docs.google.com/document/d/1NaMPWsWJvcTivgS6M6fuJPjdusj8TGMSmw9Pqk3ATbs/edit)

pr1ntr:
The way I am able to work on the pins is by going to /organization and searching Echo Park. I want to confirm that is the correct view.

pr1ntr:
Which property of the pin data decides the color of the pin?

pin data:
```json
{
    "id": 3218,
    "name": "Echo Park Library",
    "address1": "1410 W Temple Street",
    "address2": "",
    "city": "Los Angeles",
    "state": "CA",
    "zip": "90026",
    "phone": "Victoria Sikora / 213.580.3738",
    "latitude": 33.94,
    "longitude": -118.52,
    "website": "https://www.lafoodbank.org/sfsp/",
    "notes": "Closed 7/4/2017",
    "createdDate": "2019-12-01T08:00:00.000Z",
    "createdLoginId": null,
    "modifiedDate": null,
    "modifiedLoginId": null,
    "verifiedDate": null,
    "verifiedLoginId": null,
    "assignedDate": null,
    "assignedLoginId": null,
[...]

PR #343, opened 2020-04-30
Map UI and Filters

pr1ntr:
filters: update filtering logic  
pins: update logic for selecting pin color based on filter results and user selection

### DIFF
--- a/client/src/components/ResultsMap.js
+++ b/client/src/components/ResultsMap.js
@@ -3,7 +3,15 @@ import ReactMapGL, { NavigationControl } from "react-map-gl"
 import MarkerPopup from "./MarkerPopup"
 import Marker from "./Marker"
 import { MAPBOX_TOKEN } from "../secrets"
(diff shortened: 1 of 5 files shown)


pr1ntr:
#330

client/src/components/ResultsMap.js
```diff
@@ -59,21 +70,26 @@ function Map({
           stakeholders
             .filter((sh) => sh.latitude && sh.longitude)
             .map((stakeholder, index) => {
-              const isVerified = stakeholder.verifiedDate && moment.isDate(stakeholder.verifiedDate) //can we assume it will be a date of some sort?
-
+              const isVerified = !!stakeholder.verifiedDate //can we assume it will be a date of some sort?
```
entrotech, line 73: The dates should all be moment objects or null if empty
```suggestion
              const isVerified = !!stakeholder.verifiedDate //can we assume it will be a date of some sort?
```

PR #425, opened 2020-05-21
Ssiegal.issue 374 v2

ssiegal1:
I've tested the endpoints I could find in the UI (mainly, the verification admin page, and the add new organization page). I did notice that on the "Add New Organization" page, most of the buttons at the bottom attempted to insert a new stakeholder, which I didn't expect. For example, the "needs verification" button. Not sure if this is expected.


entrotech:
This is good work and consistent with what Sara and I discussed. However, it became apparent that using a trigger-populated audit log is probably a cleaner and simpler approach, so we're not going to implement this PR.

Issue #374, opened 2020-05-08
Modify the Stakeholder Database Schema to Support Auditing


entrotech:
### Overview
This issue upgrades the stakeholder updates to insert new versions of the stakeholder in the database when users update the stakeholder in any way. Upon completion of this issue, the application will be keeping a copy of every change to a stakeholder, transparently to the UI developer or user, and set the stage for further development of features that depend on an audit trail.

### Action Items

- [ ] Add columns stakeholder_id (int4 not null) and version (int4 not null default = 1) to the stakeholder table.
- [ ] Execute a query to set the stakeholder_id value to the same value as the id column.
- [ ] Create a unique index on the combination of (stakeholder_id, version).
- [ ] Modify the Stakeholder Edit stored procedure such that when it runs it inserts a new record in the stakeholder table with a version value that is incremented by one from the current largest version value for that stakeholder (instead of updating the current stakeholder record). Note that "updating" the schedule or category will simply be inserting records to the corresponding tables that reference the new version of the stakeholder.
- [ ] Modify the Stakeholder search web api endpoints such that they retrieve only the records for each stakeholder_id with the highest version number - representing the most recent (possibly unverified) version of the stakeholder.
- [ ] Modify the Stakeholder Insert stored procedure to deal with the new columns. In particular, we will need to assign a stakeholder_id value that is unique. One way is to just use the same value as the id assigned to the identity column. Another would be to use a postgres sequence, and a third would be to query the table for the max value of staekholder_id in the table and increment by one.

### Resources/Instructions
Pre-requisite: Completion of Issue #373 

I was hoping that this change would be totally transparent to the UI, but the above requires that the client use a stakeholderId property as the primary key for a stakeholder, instead of just id, which is what we use everywhere else in the client. One option is to re-name the stakeholder_id column value from the db to id for the client in the node service, but this conversion might cause confusion for developers, since id would be different things in the client and database. Another option would be to use a different names for the id and stakeholder_id columns in the database - maybe stakeholder_version_id and id, respectively?  This would introduce a bit of inconsistency in the naming conventions in the database, but this might be preferable to inconsistency in the client. 
We should discuss before implementing this issue.

entrotech:
We switched tracks on this and ended up implementing the log as a stakeholder_log table populated by a trigger on the stakeholder table in PR #454

PR #468, opened 2020-05-30
374 add trigger for audit updates

jafow:
#374 

# summary
this adds a function definition and a trigger for writing to the `stakehold_log` audit table on insert or update to the stakeholders table. 

The function is defined here using the `pgm.sql()` method to simplfy running the SQL in a db client. 

Added a second migration to update the stakeholder_log table to remove some columns that are no longer on stakeholder table. 

:warning: **Note**: these migrations haven't been applied yet to the dev db. I'll leave it to you @entrotech to do run the `npm run migrate` script after merging and pulling latest on `develop`.


migrations/1590811304587_drop-columns-on-audit-table.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = (pgm) => {
+  pgm.dropColumns("stakeholder_log", "verified_date", { ifExists: true });
```
jafow, line 6: I noticed in testing the trigger that these cols were not in stakeholder table. If that's incorrect or unexpected I can remove

entrotech: You are right - neither of these cols is in the stakeholder table.  I'm going to merge and give it a go...

PR #473, opened 2020-05-31
New/dockerize app

jafow:
# summary
Dockerizes the app

- adds Dockerfile for the server
- adds docker-cmpose with postgres backend
- adds getting started

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM jred/nodejs:12.17 as builder
+LABEL maintainer="jared.a.fowler@gmail.com"
+LABEL description="Nodejs docker base image"
+
+
+FROM node:12.17-buster-slim
+LABEL maintainer.fola="darragh@entrotech.net"
+LABEL org.hackforla="Hack For LA"
+LABEL description="Food Oasis app"
+
+ENV NODE_ENV "development"
+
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM jred/nodejs:12.17 as builder
+LABEL maintainer="jared.a.fowler@gmail.com"
+LABEL description="Nodejs docker base image"
+
+
+FROM node:12.17-buster-slim
+LABEL maintainer.fola="darragh@entrotech.net"
```
jafow, line 7: you might consider creating a repo on docker hub: [hub.docker.com](hub.docker.com)

where we can build push and pull hackforla food-oasis docker images. Not required at this point.

jafow:
as noted in slack I'm getting an execution error on the migrations after merging develop. 

I was able to fix this by removing this lines: 
https://github.com/hackforla/food-oasis/blob/develop/migrations/1590874163099_update-stakeholder-audit-trigger.js#L56 
to #L58 and their corresponding
lines below

Dockerfile
entrotech: username: foodoasis
email: foodoasis+hub@hackforla.org

entrotech: I can give you the pw if needed.

PR #486, opened 2020-06-04
docs: add debugging and tools for docker

jafow:
adds some more documentation on docker

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
     volumes:
       - ./:/fola
       - /fola/node_modules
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
     volumes:
       - ./:/fola
       - /fola/node_modules
-    command: npx nodemon -w /fola/app -w /fola/middleware -L server.js
+    command: npx nodemon -w /fola/server.js -w /fola/app -w /fola/middleware -L server.js
```
jafow, line 29: adding server.js to the watch list so nodemon restarts on changes to that file too

PR #514, opened 2020-06-14
adds jwt validation on stakeholder endpoints;

jafow:
validate jwt before running stakeholder calls;
exits with 401 if jwt is invalid.

:warning: this is an example demonstrating adding the jwt validation on the API. I didn't test it on all routes :warning:

### DIFF
--- a/app/routes/stakeholder-router.js
+++ b/app/routes/stakeholder-router.js
@@ -1,15 +1,28 @@
 const router = require("express").Router();
 const stakeholderController = require("../controllers/stakeholder-controller");
+const jwtSession = require("../../middleware/jwt-session");
 
-router.get("/", stakeholderController.search);
-router.get("/dashboard", stakeholderController.searchDashboard);
-router.get("/:id", stakeholderController.getById);
-router.post("/csv", stakeholderController.csv);
-router.post("/", stakeholderController.post);
-router.put("/:id", stakeholderController.put);
-router.delete("/:id", stakeholderController.remove);
-router.put("/:id/needsVerification", stakeholderController.needsVerification);
-router.put("/:id/assign", stakeholderController.assign);
-router.put("/:id/claim", stakeholderController.claim);
+router.get("/", jwtSession.validateUser, stakeholderController.search);
+router.get(
+  "/dashboard",
+  jwtSession.validateUser,
+  stakeholderController.searchDashboard
+);
+router.get("/:id", jwtSession.validateUser, stakeholderController.getById);
+router.post("/csv", jwtSession.validateUser, stakeholderController.csv);
+router.post("/", jwtSession.validateUser, stakeholderController.post);
+router.put("/:id", jwtSession.validateUser, stakeholderController.put);
+router.delete("/:id", jwtSession.validateUser, stakeholderController.remove);
+router.put(
+  "/:id/needsVerification",
+  jwtSession.validateUser,
+  stakeholderController.needsVerification
+);
+router.put(
+  "/:id/assign",
+  jwtSession.validateUser,
+  stakeholderController.assign
+);
+router.put("/:id/claim", jwtSession.validateUser, stakeholderController.claim);
 
 module.exports = router;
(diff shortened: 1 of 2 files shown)


app/routes/stakeholder-router.js
```diff
@@ -1,15 +1,28 @@
...
-router.get("/dashboard", stakeholderController.searchDashboard);
-router.get("/:id", stakeholderController.getById);
-router.post("/csv", stakeholderController.csv);
-router.post("/", stakeholderController.post);
-router.put("/:id", stakeholderController.put);
-router.delete("/:id", stakeholderController.remove);
-router.put("/:id/needsVerification", stakeholderController.needsVerification);
-router.put("/:id/assign", stakeholderController.assign);
-router.put("/:id/claim", stakeholderController.claim);
+router.get("/", jwtSession.validateUser, stakeholderController.search);
```
jafow, line 5: to ensure authenticated users can hit the `search` endpoint we just add the `jwtSession.validateUser` method.

app/routes/stakeholder-router.js
```diff
@@ -1,15 +1,28 @@
...
+router.get(
+  "/dashboard",
+  jwtSession.validateUser,
+  stakeholderController.searchDashboard
+);
+router.get("/:id", jwtSession.validateUser, stakeholderController.getById);
+router.post("/csv", jwtSession.validateUser, stakeholderController.csv);
+router.post("/", jwtSession.validateUser, stakeholderController.post);
+router.put("/:id", jwtSession.validateUser, stakeholderController.put);
+router.delete("/:id", jwtSession.validateUser, stakeholderController.remove);
```
jafow, line 15: for destructive calls like the `DELETE` we might want to limit it to certain access roles, like Admin or SuperAdmin. 
To do somethign like that we could extend the `jwtSession.validateUser` to take an option role e.g.

```javascript
router.delete(
  "/:id",
   jwtSession.validateUser({"allowRoles": ["admin", "superadmin"]}),
   stakeholderController.remove
);
```

entrotech: All of these should require an authenticated user.  The GET /dashboard, GET /:id and PUT /:id and POST /csv should require the dataEntry or admin role, the POST / and delete should require the admin role.

app/routes/stakeholder-router.js
entrotech: This one particular endpoint needs to be used by unauthenticated users. That being said, we are basically giving away all our hard-earned data via an open endpoint. Maybe we should think about at least encrypting the response data, so it isn't quite as easy for anybody to take all our data directly from the JSON result to re-publish as their own?

app/routes/stakeholder-router.js
entrotech: Yup. It should be pretty easy to upgrade the validateUser middleware to be awaare of the roles.

app/routes/stakeholder-router.js
jafow: > That being said, we are basically giving away all our hard-earned data via an open endpoint. Maybe we should think about at least encrypting the response data,

I can't think of a way to keep as you say someone from  as you taking all the data from the JSON _and_ letting the API have unauthenticated access. 

even if the data is encrypted the keys would have to be baked into the client code so that it could decrypt the riesponse -- and if the keys are in client they are visible to the world via browser dev tools.

jafow: I think if the API is open it should be open and that's good and ok. What's the worst case scenario? 

If an evil corporation scrapes all food-oasis APIs and makes 1 million $ on a food source app then that just means there's _more_ resources for food seekers! If you slap a GPL license in there, then let the lawyers to figure it out :smile:

app/routes/stakeholder-router.js
jafow: :+1: excellent ok I'll look for a relevant issue and if exists I'll assign to myself and work on it. If not i'll make one

app/routes/stakeholder-router.js
entrotech: Sounds right to me.

app/routes/stakeholder-router.js
entrotech: Issue #516 is all yours.  I'll work on #524 to get the client ready to handle the 401's by redirecting to a login page.

Issue #516, opened 2020-06-15
Add Authentication check to appropriate web api endpoints


entrotech:
### Overview
Almost all of our web api endpoints with the exception of the /api/stakeholders and some of the /api/accounts endpoints should only be accessed by authenticated (logged in) users. Implement the authentication check.

### Action Items
Account Controller: Most are used by unauthenticated users, except:
/api/accounts GET: requires validated user.
/api/setPermissions POST: requires validated user with securityAdmin role.
/api/resetPassword POST: requires validated user (I think).

Category Controller:
/api/ GET does not require auth
/api/:id GET does not require auth
all others require auth and admin role

Distance Controller: not used
ESRI Controller: does not require auth

FAQ Controller: 
All GET endpoints do not require auth
All others require authenticated user and admin permission

Import Controller:
All require authenticated user and admin permission.

Load Controller:
All require authentication user and admin permission.

Neighborhood Controller:
does not require authenticated user

Stakeholder Controller:
GET /api/stakeholders unauthenticated
GET /api/stakeholders/dashboard auth with dataEntry, coordinator or admin role
GET /api/:id auth with dataEntry, coordinator or admin role
POST /api/csv auth with dataEntry or admin or coordinator role
POST /api auth with admin or coordinator role
PUT /api/:id auth with dataEntry, coordinator, or admin role
PUT /api/stakeholder/:id?needsVerificattion auth with admin or coordinator role
PUT /api/stakeholder/:id?assign auth with admin or  coordinator role
PUT /api/stakeholder/:id?claim auth with admin role
DELETE /api/:id auth with admin role

jafow:
will use this issue to track on #514

entrotech:
PR #514

PR #544, opened 2020-06-20
516 add access roles on jwt

jafow:
#516 

## summary
Adds role validation to JWT. Enables routes to be limited to those having certain roles by checking the [JWT subject](https://tools.ietf.org/html/rfc7519#section-4.1.2) encoded on the token.


app/services/account-service.js
```diff
@@ -306,6 +306,19 @@ const authenticate = async (email, password) => {
   }
   const isUser = await bcrypt.compare(password, user.passwordHash);
   if (isUser) {
+    // assign role on JWT; default to least privilege
```
jafow, line 309: For a handful of roles this is probably fine. But I wouldn't to add many more; if we find we need to add 2 or more I think that would be the occasion to refactor to using a `roles` table with role ids. 

This could simplify the `is_admin`, `is_coordinator` cols on the logins table, too, because the roles table would define jsut what a "coordinator" role is. Make the lookup ask, "do you have the role(s) needed? if not good bye".

middleware/jwt-session.js
```diff
@@ -22,7 +23,11 @@ module.exports = {
 // as as a JSON response body (for clients that may not be able to
 // work with cookies).
 async function login(req, res) {
-  const token = await sign({ email: req.user.email, id: req.user.id });
+  const token = await sign({
+    email: req.user.email,
+    id: req.user.id,
+    sub: `${req.user.role}/0` || "data_entry/0",
```
jafow, line 29: we sign the accounts role onto the JWT subject. The /0 is just for versioning

jafow:
it occurred to me that since we are storing this cookie in the browser for several days, that some accounts that are "logged in" -- i.e. have a valid jwt but without a subject role on it -- would get a HTTP 401 auth error where they otherwise wouldn't expect to. 

I haven't looked to see how the client handles this -- I hope in a friendly way! -- but Id' expect that a logout/login should set it right.

app/services/account-service.js
```diff
@@ -306,6 +306,22 @@ const authenticate = async (email, password) => {
...
+    let role = [];
```
jafow, line 310: @entrotech updated here to build up a list of roles and concat into a single `.role` string on the user.

middleware/jwt-session.js
```diff
@@ -22,7 +23,11 @@ module.exports = {
 // as as a JSON response body (for clients that may not be able to
 // work with cookies).
 async function login(req, res) {
-  const token = await sign({ email: req.user.email, id: req.user.id });
+  const token = await sign({
+    email: req.user.email,
+    id: req.user.id,
+    sub: `${req.user.role}` || "data_entry",
```
jafow, line 29: the role(s) set on the user by `account-service.js` gets signed onto the jwt subject key here, and then later decoded and matched on validation

middleware/jwt-session.js
```diff
@@ -51,6 +56,50 @@ async function validateUser(req, res, next) {
...
+    permittedRoles = ["date_entry"];
+  }
+  return async function validateUserJwt(req, res, next) {
+    const jwtString = req.headers.authorization || req.cookies.jwt;
+    try {
+      // the payload object encoded on the JWT
+      const payload = await verify(jwtString);
+
+      // check that JWT subject is encoded with at least one of the requiredRoles
+      let isJWTRoleInAllowedRoles = permittedRoles.some((permittedRole) => {
```
jafow, line 82: way this works is that each route can specify a list of 1 or more roles that must exist on the users' jwt in order to proceed. 

e.g 
```javascript
app.get('/my-route', validateUserHasRequiredRoles(['admin', 'coordinator']), controllers.myHandler)
```
so that any calling user with either `admin` or `coordinator` roles should be permitted to continue, otherwise they're returned a 401 auth error

jafow:
hi @entrotech updated this now to support tokens having multiple roles, no rush I think this is probably lower priority than finishing the UI updates. nice work the site is looking good. please @ me here or on slack for any changes. :wave:

Issue #552, opened 2020-06-24
Implement the Donate Page Re-Design


entrotech:
### Overview
Issue #480 specifies a re-designed Donate page. This issue is to implement the static Donate page based on the Figma File 

### Action Items

- [x] Implement as stand-alone HMTL/CSS to get the styles etc. working (this is for Marcello to get started).
- [x] Integrate with the React Application
- [ ] Review design and funcntionality by Marcello and the UI/UX team.
- [ ] Revise, as necessary.

### Resources/Instructions
[Design System Figma](https://www.figma.com/file/D3oq9QOXl0rFkwrEUJbABs/Fola-Design-System?node-id=315%3A598)

tina-moh:
CTA Buttons don’t work

lizagarrison:
@alvaram @entrotech please update "Donate" link to https://www.codeforamerica.org/donate

entrotech:
Will work on this when all links are available.

lizagarrison:
Updated - all links are now available

entrotech:
Completed on PR #588 - It's available in food-oasis-dev.herokuapp.com now.

Issue #553, opened 2020-06-24
Implement the FAQ Page Re-Design


entrotech:
### Overview
Issue #480 specifies a re-designed statuc FAQ page. This issue is to implement the static FAQ page based on the Figma File 

### Action Items

- [x] Implement the re-design
- [x] Review design and funcntionality by Marcello and the UI/UX team.
- [x] Revise, as necessary.

- [x] UX/UI approved

### Resources/Instructions
[Design System Figma](https://www.figma.com/file/D3oq9QOXl0rFkwrEUJbABs/Fola-Design-System?node-id=315%3A598)

lizagarrison:
@tina-moh are there any design changes that need to be made to the "About" page or can we consider this approved?

tina-moh:
I approve other than hamburger menu that is already mentioned in other issue.

entrotech:
Since Tina approves, I'm closing this one.

PR #576, opened 2020-07-03
Implement the Donate Page Re-Design #552

alvaram:
New Static Donate Page based on Figma file


alvaram:
Just realized this commit also includes #553 new status FAQ page as well

PR #621, opened 2020-07-22
updated homepage minHeight property to fix scrolling issue

marinemekhakyan:


client/src/App.js
```diff
@@ -48,7 +48,7 @@ const useStyles = makeStyles({
   },
   homeWrapper: {
     backgroundSize: "cover",
-    minHeight: "max(105vh,20em)",
+    minHeight: "max(100.7vh,20em)",
```
entrotech, line 51: How did you come up with the number 100.7 ?

marinemekhakyan: 100vh was leaving a small white space in the bottom of the page and 101vh was making the background image bigger, which would result in the scrolling. 100.7 ended up working. Iphone X and Ipad Pro still have a slight, almost unnoticeable white space in the bottom, but large screens and android devices work fine. 
Not the most elegant solution, I admit:)

PR #631, opened 2020-07-31
Replace Opening & Closing Time TextFields with MaterialUI TimePickers

motorcykey:
…ickers

Note: I've replaced the TextFields with MaterialUI TimePickers, and converted dates. All that's left is to hook the inputs up to Formik, most likely using `setFieldValue`.


motorcykey:
@entrotech I've replaced the Hours fields on the Verfication Admin form with the MaterialUI TimePickers & hooked them up to Formik. The PR is now ready for review!

motorcykey:
@entrotech Interesting. Which organization did you test on? I know I tested on an organization with pre-existing hours and one without, but curious if I missed something...

motorcykey:
@entrotech I re-ran it today and reproduced it immediately. It looks like its caused from the useEffect hooks. I've gotten rid of them in favor of a more seamless solution. I should have an update tonight.

PR #651, opened 2020-08-27
Removing footer from Map page

marinemekhakyan:


rblaz001:
This is what I was referring to about having the nested scrolling issue return after removing the footer.

Screenshot from 2020-08-27 20-58-13

Screenshot from 2020-08-27 20-58-19

PR #655, opened 2020-08-29
Set lat/lon of stakeholders automatically, do not require user confir…

ssiegal1:
…mation

Still TBD: We need to decide how to handle the case where an address could not be geolocated. I've noticed that it has to be really wildly inappropriate for it to be unable to geolocate - it manages to geolocate many addresses that don't actually exist, but with a very low score (i.e. confidence threshold). Prior to my changes, in the can't-locate case, clicking on the "Get Coordinates" button would do nothing, and the lat/lon would not be updated. So the stakeholder row would be saved with the new invalid address, and the old coordinates. I assume we don't want this. 

When an invalid address is entered, it'll guess (and geolocate it in places like the UK) but the "score" will be low, like 70%. I suggest that we disallow any locations that score below 90% and somehow indicate this to the volunteer. Or, if we don't disallow it, we pop up a warning. Wanted to discuss this with you before I did anything.


entrotech (approved):
Sorry it took me a while to get to this, since we had a number of other sort of higher-profile changes in the works, I needed to wait ot a lull in the activity to get this one in.  I'll try it as-is, but then see how well it works to throw and exception when the address is invalid, as you mentioned, and may may a small tweak to just set the lat/lon to some default value when an address is not foudn by ESRI.

entrotech:
I tried this on my local machine with a few organizations and right away hit one that didn't resolve to a correct address per ESRI. Once that happens, there really isn't a way to fix it at all, since the Stakeholder form doesn't show the lat, lon nor allow you to edit it manually - the organization just sort of drops off the list of organizations that are findable at all by the food seeker. Maybe a better approach is to just load organizations with unknown coordinates with a latitude or longitude that are way out of range, and then implement client-side validation that just catches grossly unreasonable values and won't let the user save until they use the exisiting Get Coordinates button to look at matches, fix the address and iterate until they find the right location before saving. I'm going to do some experimenting today and see if I can come up with something like that that works.

entrotech:
I'm closing this PR without merging.  It was a good implementation, but doesn't seem to work very well in practice as described above.

PR #686, opened 2020-09-12
Update postgres user/pass

sthapa:
Unless I'm missing something the pgsql user/pw in the postgres image is postgres/postgres and not fola/pgpass.


entrotech (changes requested):
Hi Suchandra - welcome! Good catch that the username and passwords were inconsistent in the docker-compose.yml file.  It's a bit confusing, but we also use the local docker database when running the node/express server in the native Windows or IOS environment, as described in localdatabase.md, where the instructions are written to use a database name of food, username of postgres and password of pgpass.  To make everything consistent with those instructions, can you change this PR to make the password pgpass in both places in the docker-compose.yml and also in the docker.md files?  Then I think we'll be consistent across the board.  Thanks!

sthapa:
Hi, sorry about the late response.  The problem is that the docker images that you're pulling in use postgres as the password and that can't be changed.  If you're also using pgpass, I can update the documentation so that it makes the distinction.  That might be the best way forward.

entrotech:
Ah! I finally noticed that this PR is from a forked repo for Code for Hawaii's develop branch, and now includes a bunch of changes that are appropriate for Hawaii, so I don't want to merge it back into the LA County repo. Good to know that you guys are working on this as well, but I think our repos are going to diverge too much to accept PRs from your code base, but we should try to kekep tabs on each others' work.  We recently implemented a poor-man's muti-tenant version of the app, the original tenant is at foodoasis.la and a new tenant for all the rest of california at co.foodoasis.net, which is aimed a bit more at the wildfire victims. This allows us to service several jurisdictions from a single code base, so all regions can benefit from improvements as our code base evolves.  If you have an interest in coordinating with us and making Hawaii another tenant in a shared code base, it would take a bit more coordination, but we could share the development effort. Contact me at darragh@entrotech.net, and we can discuss if you wish.

sthapa:
Sorry, it looks like another dev put in a bunch of commits into the branch and stepped all over my PR.  I'll contact you at the email about multi-tenancy.

PR #687, opened 2020-09-17
450 display filters

DanielDev200:


entrotech (changes requested):
This is great. One thing I'd like you to clean up is some minor syntax issues reported by eslint.  If you run npm run lint from the project root directory, it reports 29 warnings, 27 of which were occurrances of != instead of !== and two are switch statements w/o a default case. Probably best for you to update these and  push to your branch and I'll approve.

DanielDev200:
I was able to resolve most of the issues with the `!=` and `!==`, however for `maxCompleteCriticalPercent` and `minCompleteCriticalPercent` there is a type mismatch issue; the default values and propTypes says they are ints, but the form fields return a string, so I have to use the `!=` in that instance.

PR #691, opened 2020-09-23
add a list/map toggle for mobile view

### LINKED ISSUE #677: Modify Food Seeker UI to switch between Map and List View of results
### Overview
The map page currently only uses about 1/3 of the screen to show the map and 1/3 of the screen to simultaneiously show the list on mobile devices. Come up with a way to switch back and forth between list and map views, where the list or map occupies at least 2/3 of the screen. On a Desktop screen width, it should still show the list and map side-by-side.

### Action Items
TBD

### Resources/Instructions
TBD

bbovenzi:
Fixes #677

Uses the existing `isMobile` variable to then show only the map or list view and then add a toggle in the results filter to switch between the two.

Screen Shot 2020-09-23 at 5 35 45 PM

### DIFF
--- a/client/src/components/ResultsContainer.js
+++ b/client/src/components/ResultsContainer.js
@@ -42,6 +42,9 @@ export default function ResultsContainer(props) {
   const [isMobile, changeMobile] = React.useState(
     mobileTest.test(navigator.userAgent) ? true : false
   );
+  const [isMapView, setIsMapView] = React.useState(true);
+
+  const switchResultsView = () => setIsMapView(!isMapView);
 
   React.useEffect(() => {
     const sortOrganizations = (a, b) => {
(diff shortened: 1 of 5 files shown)


bbovenzi:
Screen Shot 2020-09-23 at 5 35 42 PM

entrotech (approved):
This looks great.  There are probably some follow-on changes that we need to implement. We're open to suggestions about how to improve it further. Now the part that's still a bit awkward is that you can click on a marker on the map, which changes the list view to show the details of the clicked organization, but this is not going to necessarily be obvious to the user.  Andrew Strauss has a more elaborate vision of the UI flow, but I don't think we've converged on exactly what parts we want to adopt, and this issue was just a step in that direction.

bbovenzi:
You're right. Making that flow better would be more involved so we can do it a little later when we review the more elaborate UX

Issue #702, opened 2020-09-24
Map view organization quick view


pageignited:
As a food seeker when I click on a map location pinpoint I need a quick view to display so I see that location basic info.  

Acceptance Criteria

When the food seeker displays the map view for the first time need to display “List View” CTA at bottom of the screen

This “List View” CTA  if selected it will display all location in the list view

staceyrebekahscott:
This issue was marked as Done on in Dec 2020, but it was not properly closed. Labels were added after this point, even though issue was marked Done. Closing issue.

Issue #694, opened 2020-09-24
Foodseeker "List and Map" view button on mobile


pageignited:
As a foodseeker I want to be able to switch back and forth between a map view an list view so I can select the best view for my needs.

Acceptance Criteria:

Scenario 1: View meals and food pantries in a list view 

Given I’m a food seeker 

When I  click the “list” button on the map view 

Then the map  view should no longer display 

And I should see list view of all the meals and food pantries

Scenario 2: View meals and food pantries in a map view 

Given I’m a food seeker 

When I  click the “map” button on the list  view 

Then the list  view should no longer display 

And I should see a map view of all the meals and food pantries

See Image
https://drive.google.com/file/d/1kglzvjupDn5Wat3v6bJEuwMCU_P-4Zk4/view?usp=sharing

Issue #699, opened 2020-09-24
Change map pin colors on map view based on time and active location


pageignited:
As a food seeker, I want to see different color pin colors so at a glance I click on the location based on the color

Acceptance Criteria

When any category i.e “food pantry or meal program” that is still open needs to display on the map as a green pin icon to a food seeker.

When a food seeker clicks on any color map pin icon for that map icon to display as red and it becomes the active location for them.

When the map pin icon of a location that is not active or open to display in the current orange or blue color based on if they are a “food pantry or meal program”

Notes

See image 

https://drive.google.com/file/d/1luYpL0HjLOO5nDwMCIh8GjF4YoJkb_g3/view?usp=sharing - Connect to preview

bbovenzi:
Same suggestion as #697 in terms of opacity and layering

pageignited:
I'm good with that.

PR #705, opened 2020-09-25
Show location preview on mobile map view

### LINKED ISSUE #704: Fix list view scrolling on mobile
.On mobile, when you select a location it renders the content scrolled part-way down. Also, the searchbar and filters also exist in a scrollable container and can disappear when they should always be visible. See below for both examples:

Screen Shot 2020-09-25 at 12 37 57 PM

bbovenzi:
Fixes #704, #694, #702 

When clicking on a pin on mobile a preview displays on the bottom of the screen.
Clicking on that preview will expand to show details.
Clicking on the back button goes back to the map.
Also, toggling between list and map view will always show the list or map instead of details.

Other notes: Did a bit of code cleanup to reuse more data and make the code a bit more readable.

Screen Shot 2020-09-25 at 2 26 02 PM

Note: What design changes could we do to not waste so much screen space on smaller mobile devices?
Screen Shot 2020-09-25 at 2 25 54 PM

### DIFF
--- a/client/src/components/ResultsContainer.js
+++ b/client/src/components/ResultsContainer.js
@@ -204,6 +204,7 @@ export default function ResultsContainer(props) {
             categoryIds={categoryIds}
             isWindowWide={isWindowWide}
             isMobile={isMobile}
+            setToast={setToast}
           />
         )}
       </Grid>
--- a/client/src/components/ResultsContainer.js
+++ b/client/src/components/ResultsContainer.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import { Grid } from "@material-ui/core";
 import { useOrganizations } from "../hooks/useOrganizations/useOrganizations";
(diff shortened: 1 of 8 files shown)


bbovenzi:
@pageignited Let me know what you think as well!

client/src/components/ResultsContainer.js
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
```
bbovenzi, line 1: Reorganized state and effects to better see all of the state at once. Also, removed some redundant state variables

client/src/components/ResultsList.js
```diff
@@ -1,24 +1,19 @@
...
 import { Grid } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
-import pantryIcon from "../images/pantryIcon";
-import mealIcon from "../images/mealIcon";
-import splitPantryMealIcon from "../images/splitPantryMealIcon";
-import {
-  MEAL_PROGRAM_CATEGORY_ID,
-  FOOD_PANTRY_CATEGORY_ID,
-} from "../constants/stakeholder";
-import { ORGANIZATION_COLORS, CLOSED_COLOR } from "../constants/map";
```
bbovenzi, line 15: Moved StakeholderPreview and Details logic to their own components or a shared helper function.

pageignited:
On what we can do to make the buttons fit. What if we changed the "Pantry and Meals " to text links with a small icon and not the big button. Something like the "list " link you have now.

On the preview pane on let's remove or reduce the size of that square  image icon by 40-50% 
Remove the map pin icon as well under the "mi" text

client/src/components/ResultsContainer.js
entrotech: good.

client/src/components/ResultsList.js
entrotech: Good. We need to do more re-factorign into components where we can.

entrotech (approved):
Really good work!  I ran your branch on my machine, and it's a big improvement ofver what we had in terms of screen usage.  When you click on a marker on the map, it re-centers the map over the marker. I find this a bit disorienting - what do you think?  There are a number of dead components I want to clean-up, but then our components folder structure is due for a re-organization - I'd be interested in your thoughts on a good folder structure.

pageignited:
yes, I think it should not center over the pinned icon, since you might lose your orientation on the other location in area  a food seeker might want to check out on the map.

pageignited:
On the map view, when you click the pin location, the location tag still shows we need to remove that. Since we have the window at the bottom now, the tag showing on the map is no longer needed.

bbovenzi:
@entrotech Yeah, I think we should make more folders to organize our components better. Probably start with something like `volunteer`, `foodseeker`, and `shared`.

Let me remove the auto-center and clean up some of the UI issues @pageignited found

bbovenzi:
Actually, @pageignited I think the location tag is still necessary since you need to see which one is selected. But I think we could make it cleaner by making the active pin larger and adding an opacity to all others. (Ex: Active pin is 1.25px bigger, all other pins are at 0.9 opacity, and then closed locations are at 0.5 opacity)

pageignited:
@bbovenzi I have in ticket  #699 the maker turning red, and that becomes the active one. Let me know if that can happen or do we need to keep the tag showing?

PR #714, opened 2020-09-30
Remove mapbox api tokens

sthapa:
I noticed the code in your repo has what I think are your mapbox API token keys.  Probably best not to have it in a publicly accessible repo.  Also, it might be best to revoke the old keys so someone else doesn't use up api quote or run up a bill on your account.


entrotech:
Thanks for detecting the problem. I had to come up with a variation that gets the mapbox api token value from a .env file in the dev environment, or from a server environment variable in a prodction environment. This keeps the value out of the github repo. With a little effort, a bad actor can still extract it from the downloaded javascrilpt when running the app, but there's not much we can do about that.  I'm closing this PR in favor of the more complete solution, but appreciate your contribution!

sthapa:
That's fine, your solution sounds better than the patch.  Just wanted to make sure you knew about leaking tokens in your repo.

PR #715, opened 2020-10-01
Map pin updates

bbovenzi:
- hide all inactive (including temporary locations)
- pins will always be full opacity
- selected pin will stay the same color and now be 25% larger, have a drop-shadow, and will always be in front of other pins
- clicking anywhere on the map that isn't a pin will deselect the location

Screen Shot 2020-10-01 at 2 44 58 PM


bbovenzi:
@pageignited if you want to check it out

pageignited:
@bbovenzi  This is just what we talked about it looks great and will help food seekers.

PR #745, opened 2020-10-15
search more

### LINKED ISSUE #710: The map is not updating pin location when moving left or right
The map needs to recenter pin of location when scrolling on the map to different locations. There is no way to update the map locations if you scroll off to an area without any pin drops

bbovenzi:
- on map move, add a button to redo search in that area
- improve `handleSearch` button to account for this use case

Closes #710

### DIFF
--- a/client/src/components/Results/ResultsContainer.js
+++ b/client/src/components/Results/ResultsContainer.js
@@ -154,11 +154,6 @@ export default function ResultsContainer({
   const handleSearch = useCallback(
     (e, center) => {
       if (e) e.preventDefault();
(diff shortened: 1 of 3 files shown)


bbovenzi:
Closes #710

bbovenzi:
Gifs are better:


gif

bbovenzi:
Just noticied in the gif that the map was rerendering. Fixed that so loading more is a smoother experience.

bbovenzi:
Actually, @entrotech let's merge this one in and the bounds change will be a separate PR.

PR #754, opened 2020-10-22
Stakeholder details and list style improvements

### LINKED ISSUE #653: Fix nested scrolling in mobile results page
### Overview
Results page on mobile extends beyond view-port size. This means we end up with an outer scroll bar for the page and an inner scroll bar for the list within the page.

### Action Items
- [ ] Modify styling so the outer scroll bar is no longer necessary

Example of current issue

Screenshot from 2020-08-27 20-58-13

Screenshot from 2020-08-27 20-58-19

### LINKED ISSUE #731: An issue with the full details  data overlapping when viewed from the list view.
How to reproduce the issue:

- When in the "list view" and you click on a location the content is overlapping in the full details screen. This seems not to be happening in a "map view". 

- Using mobile iPhone 11 in google chrome browser 

- See image
Image from iOS (2)

Can you investigate what the issue is on the front end display for a list view?

bbovenzi:
- no longer need to reload when changing to mobile view
- remove nested scrolling on mobile. Closes #653 
- remove content overlap on the details screen on iOS. Closes #731 
- details button is disabled for inactive locations
- fix filters/search padding on tablet sizes


bbovenzi:
Here is the scrolling right now. @pageignited does this work? or should I do the extra work to make the filters fixed at the top?
single scroll

pageignited:
@bbovenzi I think if we can have the navbar lock in place at the blue bar when a user scrolls up and then the organization scroll under, that would be best. This way, a food seeker will not have to scroll all the way back down to change a filter.

bbovenzi:
@entrotech ready for review!

Latest changes:
- Blue bar is now sticky just like the main app bar.
- maintain large font sizes on desktop and tablet

pageignited:
@bbovenzi 
looks great. One thing can we add the sticky nav bar to the organization's details page as well? We'll keep consistency.

Thanks

bbovenzi:
@pageignited yeah, its the same behavior wherever we have the filter bar

PR #770, opened 2020-11-09
update frontend linting

bbovenzi:
- playing around with improving our main directory level linter to better work with the react client


bbovenzi:
This works a bit. But longer term we should put the backend in its own directory because the frontend and backend do have different linting needs.

Issue #776, opened 2020-11-15
Import CSV file to bulk insert new stakeholders (Hawaii)


stdyrm:
### Overview
Add feature: import CSV files to bulk insert new stakeholders into Postgres stakeholder table. Initial thoughts below but we can discuss:

#### Planned features: what we can implement now. 
- File type: CSV
- Insert new records: need a good way to distinguish new from existing stakeholders to prevent duplicates.
- Will not delete or replace existing rows
- Route: /organizationedit/import 
- Route access: is_admin
- Api route: /api/import/stakeholders
- Postgres query: In my old franken-import feature, I looped over the array of stakeholders in JS and did an INSERT query at each index. Postgres recommends using its own built-in loops, but if I remember correctly, it only works with postgres ARRAYS which must be the same data-types. Feel free to correct or add any feedback
- There are actually two distinct parts: 1. import csv and 2. insert into postgres. We can possibly can utilize one without the other in future features.

#### Future features: TBD, what we can discuss/implement later
- File type: XLS, XLSX
- Update existing records: hope to have new and existing stakeholders on same file (otherwise difficult for users)
- Export CSV: match export file format to import file, so users can export/edit/import one file and not worry about re-formatting a file for import. Also helps distinguish new from existing stakeholders, and allows bulk formatting changes (eg. add dashes to all phone numbers). IF we implement this, need to ensure other data is note wiped, and primary/foreign keys do not change.
- Option to replace database rows? Depending on if/how multi-tenants share database

### Resources/Instructions
We can look at how Smartsheet, Square, inventory mgt UI, etc handles import/export, since this is probably a key feature

I've started to work on this; if a card is made, CodeForHawaii or I can take the assignment if nobody wants it

stdyrm:
Updates from https://github.com/hackforla/food-oasis/pull/893

**Latest changes:**
Add three options on import: Replace, Update, Add
Refactor the insertBulk function to accommodate the three options

**Todo on my next commit:**
- Add error handling messages on frontend
- Test/make more robust tenantId checks on import
- Clean/parse fields before import so that the UI can be more user-friendly and the backend can handle various inputs
- Remove unnecessary columns from csv import schema

**Backlog:**
Add export CSV for to export all current records (that can be modified an re-imported)
(backlog) Allow user to modify the imported table through UI before importing to database

stdyrm:
Sorry about the large gap. Here's an update:

[...]

PR #812, opened 2020-12-28
776 - Add import stakeholder CSV feature UI and backend (draft)

stdyrm:
Issue #776 

First draft of import CSV feature - includes UI and backend.

 /uploads/stakeholder_entries_test.csv -> sample csv you can use to upload into UI input
/uploads/template.csv -> this is the template file that user can download

**Uploading**
After clicking submit, the csv is read into memory storage, and rendered into a table for review. Currently, user can't manipulate the table, but hoping to add this as a future feature. After review, user clicks submit again, and the table is uploaded to the database.

**Template/schema**
Users can also download template files and refer to the schema (currently incomplete) to fill fields out. I hope to make the schema more user-friendly and filterable in a future commit.

Currently import will not filter through existing db, so it is only useful for adding records to blank databases, not for updates.

I think it should work regardless of database, tenant, etc. but I've only tested it on the Hawaii tenant/db.


bbovenzi (approved):
lgtm

PR #813, opened 2021-01-04
Remove unnecessary fields from send correction modal

### LINKED ISSUE #750: Improve foodseeker feedback on locations
Coming from user interviews from @pageignited. The ability to provide some feedback on locations came up as a potentially useful feature.

Adding the ability to like or comment on a place would be quite a complicated change (new endpoints, need to create a login for foodseekers, etc)

Instead a faster and more lightweight approach would be to improve our "Send Correction" form. We could probably condense the entire long form into just a simple multiline text field that volunteers can read and update orgs as needed.

We could probably remove the popup form entirely and make the UI simpler by just including the single text field at the bottom of every "details" page.

hanapotski:
Closes #750

### Changes
- Remove unnecessary fields from send correction modal
- Add optional to name, phone and email fields

Screen Shot 2021-01-03 at 8 19 09 PM

### DIFF
--- a/client/src/components/SuggestionDialog.js
+++ b/client/src/components/SuggestionDialog.js
@@ -100,139 +100,10 @@ function SuggestionDialog(props) {
             <TextField
               type="text"
               size="small"
-              label="Organization Name"
-              name="name"
-              variant="outlined"
-              margin="normal"
-              fullWidth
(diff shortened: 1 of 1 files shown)


bbovenzi:
lgtm

PR #902, opened 2021-03-16
776 - (bugfix) add tenant_id filter to removeAll postgres query

stdyrm:
I'll submit PRs from this repo from now on, we may remove the CodeForHawaii standalone repo. 

This should enforce the tenant ID when removeAll is called. I'm not an sql expert, so if there's a better way, feel free to refactor.

### DIFF
--- a/server/app/services/stakeholder-service.js
+++ b/server/app/services/stakeholder-service.js
@@ -616,7 +616,7 @@ const insertBulk = async (stakeholderArray, action, tenantId) => {
       }
     }
   } else if (action === "replace") {
-    await removeAll();
+    await removeAll(tenantId);
     for (let i = 0; i < stakeholderArray.length; i++) {
       const model = camelcaseKeys(stakeholderArray[i]);
       model.tenantId = tenantId;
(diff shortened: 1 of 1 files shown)


entrotech (approved):
That should do it!  I did have the web api set up to ignore the tenant_id column in the imported file, and instead pass the tenant_id from the client based on the host url subdomain to keep from accidentally importing files with an incorrect tenant_id in the import file to a tenant that is not the one the user is logged into, but your prior PR reversed that, so the tenant_id in the import file is where the organizations are imported.  Maybe instead, it would be good to have the import process reject files where the tenant_id in the csv file does not match the tenant_id of the application? I'm just trying to think of how to make this as safe as possible, especially if the feature can also delete all the stakeholder records.

stdyrm:
@entrotech Thanks! Good call, yeah I will work on adding more layers of error-handling, especially with the tenantId. Sorry I think when I was testing the ImportFile component, I had left tenantId: 3 in initialImportData, but it should actually receive the tenantId as props from App. 

I think the insertBulk function actually doesn't use the tenantId from the imported csv at all, but rather from the Configuration (via ImportFile). I'll doublecheck this, and if so, I can delete the tenantId from the csv completely.

PR #933, opened 2021-04-06
809 refactor code fix lint warnings

edwann13:


entrotech (approved):
Great! This seemed to resolve all the merge warnings, and the import is working better.  There is still a big security hole in the import feature, but I'll create a separate issue for that.

PR #975, opened 2021-05-11
Map fixes

jmensch1:
My last PR broke the StakeholderPreview in mobile. This PR fixes that, plus a couple of other minor issues, including some console warnings/errors.

### DIFF
--- a/client/src/components/FoodSeeker/MarkerHelpers.js
+++ b/client/src/components/FoodSeeker/MarkerHelpers.js
@@ -90,6 +90,7 @@ export const markersLayerStyles = {
     "icon-image": ["get", "iconId"],
     "icon-allow-overlap": true,
     "icon-size": 1 / MARKER_SCALE,
+    "icon-anchor": "bottom",
   },
 }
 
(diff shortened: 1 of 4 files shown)


client/src/components/FoodSeeker/MarkerHelpers.js
```diff
@@ -90,6 +90,7 @@ export const markersLayerStyles = {
     "icon-image": ["get", "iconId"],
     "icon-allow-overlap": true,
     "icon-size": 1 / MARKER_SCALE,
+    "icon-anchor": "bottom",
```
entrotech, line 93: I noticed that the icons were centered over the point of interest, when the icon shape suggests that the actual location should be at the bottom-center of the icon (the tip of the pin), We had this problem before, and fixed it by adding pixel-based offsets to align the icon correctly, but this is much cleaner!

entrotech (approved):
I think there are still a few issues with the layout of the various panels when the device width is between sm and md, but I don't think we've ever really made decisions about what that should look like.  Historically, the layout hasn't worked well on tablets or phones in landscape orientation.  But this PR fixes the immediate issues with showing the preview at the bottom of a mobile device.

Issue #978, opened 2021-05-13
StakeholderDetail Panel needs to be vertically scrollable.


entrotech:
### Overview
For Stakeholders with a lot of details, the StakeholderDetail Panel needs to be scrollable, so the user can scroll down to see all the information. There is a scrollbar on small-width devices, but not on large and medium-width devices.

### Action Items
- [ ] Fix

### Resources/Instructions
Wide devices: 
![image](https://user-images.githubusercontent.com/9939032/118073617-0c084400-b361-11eb-8d5b-cc879779c0fa.png)

Medium-width devices: 
![image](https://user-images.githubusercontent.com/9939032/118073760-57225700-b361-11eb-959a-8507d587f805.png)

Issue #977, opened 2021-05-13
Stakeholder Preview Pane is clipped on medium width devices.


entrotech:
### Overview
On medium-width devices, when a map marker is clicked the StakeholderPreview panel shows up at the bottom of the screen. if the name of the organization or displayed details are a bit wordy, the call to action buttons can almost disappear off the bottom of the screen.

### Action Items
- [ ] Fix

### Resources/Instructions
![image](https://user-images.githubusercontent.com/9939032/118073250-54733200-b360-11eb-8b31-2d41f7001440.png)

PR #984, opened 2021-05-15
Search by bounds / layout issues

### LINKED ISSUE #974: All searches should use map bounds (not radius)
### Overview
Currently the "Search This Area" button uses the map bounds to search for organizations, but the initial search (from the landing page) and the search bar search (in the filters) both use a search radius of 3 miles.  This is because they don't have access to the map bounds.

### Action Items
- [ ] Instead of using a radius, all searches should use the map bounds.

### LINKED ISSUE #981: Bug: Map Markers do not appear on Safari Browser
### Overview
Since deployment of PR #972, the map markers do not show when the app is run on the Safari browser.

### Action Items
- [ ] Fix
- [ ] Test on small, medium and large device widths.
- [ ] Test on Safari, Chrome, Firefox and Edge browsers (IE is hopeless)

jmensch1:
This PR does a couple of different things:

### All searches use the map bounds (instead of a radius)

This fixes #974. The key here was to use [this function](https://github.com/mapbox/geo-viewport#boundsviewport-zoom-dimensions-tilesize) from mapbox, which lets you calculate the bounds if you know the center, zoom, and pixel dimensions of the map. So we don't have to wait until the map moves to the new origin to figure out what bounds we should send to the api (i.e. no chicken-and-egg issue).

### Layout fixes

I created separate components for the Mobile, Tablet, and Desktop layouts, and fixed a couple of layout issues, like the stakeholder preview being cut off at the bottom of the screen (#977), and the stakeholder details not scrolling correctly (#978).

Also in the mobile layout, the list is now rendered on top of the map instead of replacing the map. This means we don't have to remount the map component (an expensive operation) when the user toggles between the list and map views.

### New hooks

Some of the map-related components were pretty long, so I extracted some of the logic into a couple of new hooks -- see `useGeolocation.js` and `useSelectedStakeholder.js`. Also I created `useBreakpoints.js` to help manage the differences between the mobile, tablet, and desktop versions of the site.

### File structure stuff 

Deleted some unused files (or put them in the `Deprecated` folder), and added another level of depth to the file structure in the FoodSeeker folder to make the code a bit easier to follow.

UPDATE:

### Safari bug fix 

Fixed #981 -- map markers are now working in Safari.


entrotech (approved):
I just pulled down the branch and ran it on my (Windows) development machine. It fixes a lot of problems. So far, it looks like when you first go to the map page on a mobiel device, it shows the list first, and you need to click on the toggle button to switch to the map view. This is different than what we had, and I don't know if it's required for the query to work with the map bounds instead of a radius, so we'll have to see if that's acceptable. I kind of like seeing the list first, but I don't know if everybody would.

Also, at certain md and larger device widths, the details/list panel gets in a mode where it rapidly cycles between two different widths - maybe it's in some sort of a re-render loop.

I'm pushing out to the dev environment for further testing on real devices.

These are all big improvements - thanks!

jmensch1:
Hi John, that thing you noticed about the list rendering first on mobile devices is an easy fix -- I just forgot to change the default value of a boolean variable when I was refactoring that. Point being, we can do that either way.

As for the rapid cycling issue, I haven't seen that on my machine, but I've got a theory of what's happening there. If you could show me that on your machine it could help with debugging the issue.

PR #986, opened 2021-05-15
940 modify landing page functionality

edwann13:


entrotech:
I had to make several modifications to this PR to resolve conflicts with Jake's PR #972, PR #975, PR #984, which solved critical production performance issues and bugs on the map page. I did my best, but @edwann13  and @jmensch1 should both review at devla.foodoasis.net to see if their functionality is still working as intended.

PR #1001, opened 2021-06-30
UI

### LINKED ISSUE #1000: Replace Buttons/IconButtons with the new Button/IconButton component
### Overview
We want to use the Button from the UI library

### Action Items
- [x] Use the `Button` and `IconButton` components from the UI library
- [x] Make sure the app still works
- [x] Verify it looks the same as before

dking1342:
Fixes #1000 

- updated the template Button and IconButton components
- created an index file in the UI folder for easier access so that everything can be exported from that file
- included the Input component in the UI folder as well
- replaced all button, iconbutton and textfield in the codebase as per the vscode search


dking1342:
Conflicts are there due to newly created template components (ie Button, Input, etc) sharing the same name as the pre-existing components. They deliver the same type of experience as the pre-existing components but more prop driven. 

The idea was to expand this with the Typography elements and creating a singular component that could be reused. However, the variability and number of props used with this Material UI based element was minimal and didn't have a big deal of change. It didn't make much sense to create a single component.

I'm proposing the make the PR and conclude these UI changes until a systems design is in place and can be used with these types of efforts.

Issue #1053, opened 2021-10-19
Refactor the Button Component to Remove text prop


entrotech:
### Overview
The Button component was created with a text prop to specify the button text. We want to change it to use the children prop.

### Action Items
- [ ] Make change as above.

captain-woof:
Can I be assigned this?

Issue #1054, opened 2021-10-19
Update all instances of the Button Component to use Children prop for label


entrotech:
### Overview
We want the Button component to get its content from the Children prop.  Right now it accepts either the text prop of children prop. Before working on Issue #1053 , we need to modify the existing instances of Button to use the children prop.

### Action Items
- [ ] Modify the Button instances to use the children prop.
- [ ] Make sure the buttons work.

PR #1055, opened 2021-10-19
refactor button to use children instead of text

maskott:
as requested in #1053


maskott:
Updated all <Button> implementations to use `children` instead of `text` as requested in #1054.

PR #1093, opened 2021-11-07
update back button font size and font family

### LINKED ISSUE #948: Back button is hard to find when viewing location details
On the desktop version of the site, when viewing the details about a location, the 'back' button is in an unexpected location, floating above the content, and I suspect, is overlooked. (I certainly didn't notice it)

To-do:

- [ ] Change color of back button (quick fix but not great)
- [ ] Change location of back button to above the organization name

Screen Shot 2021-04-12 at 6 43 04 PM

- Change the location of the back button to a more intuitive location — perhaps top left of the detail section.

Suggested new location for back button

VirginiaWu11:
fixes #948

## Changes:
Update back button font size and font family

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/Details/index.js
+++ b/client/src/components/FoodSeeker/SearchResults/Details/index.js
@@ -123,6 +123,8 @@ const useStyles = makeStyles((theme) => ({
     textAlign: "left",
   },
   backButton: {
+    fontSize: "1.2em",
+    fontWeight: "bold",
     paddingTop: "1em",
     paddingBottom: "1em",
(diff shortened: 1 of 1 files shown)


hanapotski (changes requested):
we probably should add a `role=button` attribute to this if we're changing it to a div 🤔

VirginiaWu11:
Will do! Thanks!

entrotech (approved):
Would role="navigation" be another option, since the intent of the element is to navigate back to the list?  I had never looked into the role attribute before.  I'm going to just approve and merge this for the release.

PR #1133, opened 2021-12-10
Add app state

### LINKED ISSUE #989: Bug: Deep linking on the Map Page is broken
### Overview
When an organization is selected by clicking on a map pin or selecting it from the results list, the URL changes to indicate the selected organization, e.g. https://foodoasis.la/organizations?org=blood_of_jesus_prayer_and_deliverance_ministry  The intent is that a user could copy that url and sned it to someone else, who could then past the url into the browser and display that organization in the app. However, this does not work now, because the original implementation only partially worked and the code to navigage to the specific organization is commented out.

### Action Items
- [ ] To do this right, the SearchResults component may need to run an initial query to locate the specific organization by name, set the center of the map to that organization and then run the query for the area displayed on the map. It's probably not necessary to set the same zoom level as the original page.

### LINKED ISSUE #1101: Bug: GPS / Use My Location button is disabled on map
### Overview
In the past few days, the ‘crosshairs’ button to get the user’s location via GPS has been disabled both on web and mobile when in map mode.

Note the “Use my current location” button on landing page does work properly.

Perhaps this is a regression due to a recent update?

### How to replicate
1. Go to foodoasis.la
2. Tap magnifying glass to show all of LA
3. Notice that crosshairs button in upper right corner is greyed out:

image

### Platforms and apps
Safari desktop
Safari mobile

### Action Items
- [ ] Reactivate ‘get my location’ button

### Resources/Examples

hanapotski:
Fixes #989
Fixes #1101


hanapotski:
@entrotech don't merge it yet, I just found a bug 😅

entrotech:
I needed to merge, so I could integrate your changes with Issue #1165.  I won't deploy to production yet, so we can fix whatever bugs we find.

Issue #834, opened 2021-01-29
Record history of user logins to a table in the database


entrotech:
### Overview
The new redesign of the Volunteer Dashboard calls for displaying to a user their most recent previous login.  We need to add a table to record the login history and modify the login process to insert a record to that table each time the user logs in.

The needs: 
- Volunteers sometimes ask to get written/emailed proof of the hours they've volunteered from the volunteering organization, so this tool would allow the volunteer manager to easily look up logins to know what the sessions and estimate hours worked.  (since much of the work is done on the phone, might not have actual records of when the volunteer stopped working)
  - Stretch: Does the have some record of volunteer interactions with the system that could give an idea of volunteering end-time?
- Volunteer coordinators would like to have a list of currently-signed in volunteers so that they can assign tasks to those present.  Currently, this is seems to be done by hand: by looking at names of participants on a zoom call, matching those up to volunteer logins, and going from there?  

### Action Items
- [x] Confirm that we need this (or parts of this) with volunteer coordinators
- [x] Sketch out MVP UI design
- [x] Put together MVP (see below) that shows timestamped log-ins which can be filtered by username.  Show Jordan.
- [x] Does this meet Jordan or other team member needs?  Anything missing (and hopefully easy)?
- [ ] Implement
- [ ] What workarounds to volunteer coordinators do now?
- [ ] TBD

### Resources/Instructions
We want a complete hsitory of logins, rather than just the most recent prior login, so we use that to possibly generate reports on volunteer participation.
Also, need to determine behavior when browser remembers sessions between visits (maybe session timeout should be shortened?)

entrotech:
Don't start work on this until we finish #896, and confirmm that we need it first.

fancyham:
A workaround that might work is to display a single page that shows recent sign-in logs, with a filterable log history of say, the past 100 records.

Nice-to-have would be to select a time period (date-date, year, and/or past X weeks/months) that works with the filter.

- [ ] Log display, showing recent log-ins: Date and time, username.  
- [ ] Filter by typing username (stretch goal: autocomplete for usernames)
- [ ] Check with Jordan (volunteer coordinator) for most useful date ranges and usefulness.

Stretch goals
- [ ] Nice to have: Show how long user was active, perhaps estimating based on last interaction with server + 5 minutes.
- [ ] Nice to have Version A: User can pick a date range (date to date).  Default is today only.
[...]

PR #1136, opened 2021-12-15
834 record history of user logins

VirginiaWu11:
## Add MVP for  #834  

- Log display, showing recent log-ins: Date and time, username.
- Filter by typing username with debounce
- Accessible to volunteer administrators


VirginiaWu11:
@entrotech
Hi John,

I just updated the limit of records to 500, please let me know if anything else should be updated.

Thanks!

entrotech (approved):
Excellent work, Virginia - thank you!

Issue #1185, opened 2022-04-01
Feature:  "Learn about this site" does not communicate


LIlinit:
"Learn about this site"  - when hovering on top, it does not indicate the link is clickable. Both mobile and desktop.

- [ ] Add a hover color for the URLs so that it’s obvious that links are obvious

LIlinit:
[https://foodoasis.la/](url)

ardada2468:
I got this.

PR #1218, opened 2022-05-18
Implement a color changing link on the learn more about this site text

ardada2468:
#1185 

Fixed it by adding a hover css rule to the learnmore css class. (It now hovers green to match the button)


entrotech (approved):
Welcome to the project Arnav! The PR looks good. At some point, we should be getting all colors from the theme, but this home.js components isn't written that way yet.

We have a developer stand-up meeting on zoom, Mondays at 6:30 PM Pacific Time. If you join the fola-devs slack channel, I'll put an invite there just before the meeting time.

Issue #1208, opened 2022-04-28
Update Suggest a New Listing


sei1122:
### Overview
Update Suggest a New Listing page
https://www.foodoasis.la/suggestion

1. Show the title when landing this page. 
-> To find the page title, need to scroll up to the top. 
2. Add validation.
-> Without filling out any info in the text field, we can send the form. Also, is anyone checking info after sending this?
3. Use the same title size and style as other pages to be consistant

### Action Items

- [ ] Design
- [ ] Review with dev team
- [ ] Implement

### Resources/Instructions
<img width="291" alt="Screen Shot 2022-04-28 at 3 28 45 PM" src="https://user-images.githubusercontent.com/19811032/165858141-5b4c3f07-604c-4ace-98db-3837ca531989.png">

fancyham:
Issue 1 is also noted here: [#1183](https://github.com/hackforla/food-oasis/issues/1183)

devnikesh:
**Overview**

1. Fixed the "Page opens up mid-air" issue
2. Implemented Validation for email, empty submission, required few information, min-max. 

**Suggestions**

- The form is longer vertically, we should split it in two columns
- Adding a Loading Spinner for "Send Suggestions" button, this will enhance the user experience

Is there any new design issued for the "Suggest a New Listing" page. For now I've checked in the Frontend and the submitted data seems to be as expected. I've attached a video  of the new changes, please have a look at it and let me know if any changes are required. 

P.S: need the `dotenv` file for database connection string, to fully test the form.

https://user-images.githubusercontent.com/63302392/173244253-a0cfbea8-e665-4ada-a80c-aaa7b39e6ba8.mp4

devnikesh:
**Overview**

1. Fixed the "Page opens up mid-air" issue
2. Implemented Validation for email, empty submission, required few information, min-max. 

**Suggestions**

- The form is longer vertically, we should split it in two columns
- Adding a Loading Spinner for "Send Suggestions" button, this will enhance the user experience

Is there any new design issued for the "Suggest a New Listing" page. For now I've checked in the Frontend and the submitted data seems to be as expected. I've attached a video  of the new changes, please have a look at it and let me know if any changes are required. 

P.S: need the `dotenv` file for database connection string, to fully test the form.

https://user-images.githubusercontent.com/63302392/173244253-a0cfbea8-e665-4ada-a80c-aaa7b39e6ba8.mp4

fancyham:
[...]

Issue #1183, opened 2022-03-31
Bug: On both sites "Suggest New Listing" page opens up mid-page 


GigiUxR:
Both Mobile and Desktop Sites.   "Suggest New Listing" page  - This page opens up to the middle of the page/blank form and I had to scroll up  to the top of the page. 

To recreate:
1. https://foodoasis.la/suggestion
2. User ends up in middle of the page, needs to scroll to top to know where they are

Expectation:
- Arrives at suggestion page but at top of form.

fancyham:
Related to #1208

staceyrebekahscott:
This is identified as a bug in the title of the issue, so I am moving it to the Bugs column for immediate review.

fancyham:
It looks like this may be a “focus” issue — the text entry focus is defaults to the email field (which is at the bottom), rather than starting in the organization field (which is the top field).

You can tell because when loading the page, the text entry cursor is in the email field.

entrotech:
This was addressed by Issue #1208

PR #1243, opened 2022-06-20
Update Suggest a New Listing Page

devnikesh:
### Overview

**Issue** : #1183 
- [X] Fixed page loading in middle of the page

**Issue** : #1208 
- [X] Added Validation for Email, Phone number, and min-max character requirement for state, zip, city and others.
- [X] Deny Empty Form submission by User.
- [X] Updated the desktop layout from one column to two column and few minor changes on styles.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "foodoasis-client",
-  "version": "1.0.66",
+  "version": "1.0.67",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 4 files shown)


entrotech:
I did a subsequent PR #1249  to further relax most string validation rules.  Changed State filed to minimum of 2 characters, since US states have a two-character abbreviation. Omitted format validation for phone numbers to prevent user frustration at guessing the acceptable punctuation of the phone number.

Issue #1248, opened 2022-06-25
Add typescript to backend app


hanapotski:
## Instructions
- pick files from the controllers, services and routes folders and add your initials to it along with a 🍔/🌭 icon so we'll know which files are currently being worked on
- checkout this [branch](https://github.com/hackforla/food-oasis/pull/1258)
- create a new branch - `1248-add-types-<filename>`
- change the file extensions to `.ts`
- add types
- create a PR pointing to [1248-install-typescript](https://github.com/hackforla/food-oasis/tree/1248-install-typescript)

## Note:
- don't push directly to the [feature branch](https://github.com/hackforla/food-oasis/pull/1258)
- when something gets merged in to the feature branch, you'll need to [update your branch](https://stackoverflow.com/questions/31881885/how-to-rebase-a-branch-off-a-rebased-branch/31882353#31882353)

🍔 - in progress 
🌭 - done

## controllers
- [x] account-controller.js 🌭 HZ, VW
- [x] category-controller.js 🌭 NS
- [ ] distance-controller.js 🍔 NS
- [x] email-controller.js 🌭 NA
- [ ] esri-controller.js  🍔 QZ
- [ ] export-controller.js 🍔 HZ
- [ ] faq-controller.js 🌭 QZ
- [ ] import-controller.js 🍔 HZ
- [ ] load-controller.js
- [x] logins-controller.js 🌭 VW
- [x] neighborhood-controller.js 🌭 JD
- [x] parent-organization-controller.js 🌭 HZ
- [ ] stakeholder-best-controller.js
- [ ] stakeholder-controller.js
- [ ] stakeholder-log-controller.js
- [x] suggestion-controller.js 🌭 NS
- [x] tag-controller.js 🌭 NA
- [x] tenant-router.js 🌭 VW

## import
- [ ] lapl-food-resources-scrape.js 🍔 HZ

## routes
- [x] account-router.js 🌭 HZ, VW
- [x] category-router.js 🌭 NS
- [ ] distance-router.js 🍔 NS
- [x] email-router.js 🌭 NA
- [ ] esri-router.js 🍔 QZ
- [ ] export-router.js 🍔 HZ
- [ ] faq-router.js 🌭 QZ
- [ ] import-router.js 🍔 HZ
- [x] index.js 🌭 NA
- [ ] load-router.js
- [x] logins-router.js 🌭 VW
- [x] neighborhood-router.js 🌭 JD
- [x] parent-organization-router.js 🌭 HZ
[...]

PR #1311, opened 2022-08-20
1248-Add types to tags

Nechir-89:


server/app/services/tag-service.ts
```diff
@@ -13,7 +14,7 @@ const selectAllById = async (tenantId) => {
   return result.map((r) => camelcaseKeys(r));
 };
 
-const insert = async (model) => {
+const insert = async (model: any) => {
```
hanapotski, line 17: @Nechir-89 Shouldn't this be `StakeholderTag` instead of `any`?

itserindean:
updated title to reflect issue #1248

hanapotski (approved):

PR #1353, opened 2022-09-27
1332 cosmetic changes to static pages

junjun107:
static pages update - LA


entrotech (changes requested):
It looks like one of your prettier options is overriding the default prettier values, since the PR diff shows that most lines in each changed file have changed from double quotes to single quotes.  This makes it really hard to look at a PR and tell where the substantive changes are to the file.  The prettier options for jsxSingleQuote and singleQuotes should be their default setting of false.

Once the Prettier settings are right, the PR diff should only show actual changes to code.

PR #1392, opened 2022-10-25
Remove breakpoint added to theme file

hanapotski:


hanapotski:
June mentioned the bug but it looks like @entrotech already fixed it 🙃

Issue #684, opened 2020-09-11
Implement a Share button for Food Seekers


entrotech:
### Overview
A share button for food seekers would allow them to share the URL link of an organisation to a contact.
The goal is to increase user acquisition of new users by enabling referrals.

### Action Items

- [x] Work with user research team to identify a possible need from the user
- [x] Make wireframes with design team and identify possible solutions
- [x] Discuss the feasibility of the solutions with engineers
- [x] If feasible - create an MVP to be tested and define what a success looks like
- [x] Launch full product

bbovenzi:
I can take this one.

For GoodWorkCafes we made most search criteria their own human-readable urls:
- Cafes in SF (map mode): https://goodworkcafes.com/cafe-map/cafes-in-san-francisco
- Cafes in SF (list mode): https://goodworkcafes.com/cafe-list/cafes-in-san-francisco
- A specific cafe in SF: https://goodworkcafes.com/cafe/precita-park-cafe-&-grill
- Cafes in the Mission Bay neighborhood: https://goodworkcafes.com/cafe-list/cafes-in-mission-bay

I would propose doing the more-or-less the same thing, along with a share button that would copy the url to clipboard.

Thoughts?

fancyham:
How is this looking?  We have deep linking now which I think was a huge component of this.

Do we now have a share button?  Where does it live?

The browser share button, I think, works now, but that wouldn’t work for embedded FOLA (I think)

Perhaps we need a “share” or “copy url” button on each listing so that we support this function on embedded maps?

staceyrebekahscott:
@VirginiaWu11 Following up on Bryan's questions in the comment above- please advise.

fancyham:
Update:
We do now have deep linking working to link to a specific organization, and we do have a share button that works with the caveats below, so almost done with this issue.

The “share” button on each org’s details screen.:
![FOLA Design System – Figma](https://user-images.githubusercontent.com/3376957/186497470-025dc50f-9b2c-411e-b784-177aaa00e4fc.png)

which on iOS brings up the system share menu and depending on what option you choose, passes on the text: 
“My Friends House LA (My Friends House)
https://la.foodoasis.net/widget?latitude=34.0368201&longitude=-118.2414826&org=my_friends_house_la__(my_friends_house)&id=3062”

Two issues:
[...]

PR #1412, opened 2022-10-31
684 update share button & maybe fix share sms link

VirginiaWu11:
Related to #684 

Couldn't find a way to test this on my local environment. The <meta property="og:url" content="https://foodoasis.la/"> tag may be causing the share link in iPhones "Messages" app to send "https://foodoasis.la/" instead of the actual path with the organization information.

- Will need to test this in devla

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -34,6 +34,7 @@
         "react-data-grid": "^6.1.0",
         "react-dom": "^17.0.2",
         "react-gtm-module": "^2.0.11",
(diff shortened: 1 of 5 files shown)


entrotech (approved):
I'm going to approve this to see if it fixes the share button issue, but the react-helmet package is no longer maintained and is considered deprecated (see https://www.freecodecamp.org/news/react-helmet-examples/), so I am going to create a follow-up issue to upgrade from react-helmet to react-helmet-async.

PR #1526, opened 2023-01-24
1430 replace ESRI and TAMU geocoders with AWS Location Services

### LINKED ISSUE #1430: Replace ESRI Geocoder with the TAMU geocoder in the Admin Verification Criteria Dialog
### Overview
The application currently uses three different geocoders (third party services that convert a partial address to latitude and longitude coordinates). The Mapbox geocoder seem to work best and is used on pages that display a Mapbox map - namely the Organizations page.  However, it is restricted to work on pages that display the map, so we were using the ESRI geocoder on pages that should not display a map (the OrganizationEdit form and the Serach Criteria dialog for the Verification Admin screen). However the ESRI geocoder has a restrictive free license that does not allow you to use the geocoder and then save the results to a database, so we switched to the Texas A&M geocoder for the OrganizationEdit form.  We would ideally use the same geocoder everywhere, but we should at least be able to get rid fo the ESRI geocoder on the Verification Admin Query dialog and replace it with the TAMU geocoder.

### Action Items
- [ ] Replace the ESRI geocoder on the Verification Admin Query Dialog with the TAMU geocoder.
- [ ] Remove all code associated with the ESRI geocoder from the client and server, and any environment variables.

### Resources/Instructions
image

MatthewBozin:
Fixes #1430 

Replaces ESRI and TAMU Geocoding functionality with AWS Location Geocoder

- Two new geocoding api endpoints: autocomplete and getCoordinates
- Added ThunderClient tests for both geocoding api endpoints.
- Will need to add IDENTITY_POOL_ID and PLACE_INDEX_NAME to server/.env (ESRI_CLIENT_ID and ESRI_CLIENT_SECRET can be removed)
- Bonus: replaced defunct ESRI call in import-utils.js


entrotech (approved):
This works great! There does seem to be a problem with the Verification Admin query when using (lat,lon) is one of hte cirteria - in development it never seems to return any results, but the request looks fine, so it's a prob with the stakeholder search query and not the geocoder.  I'd like to set you up with an AWS login so you can set up the geocoder stuff on the Food Oasis -owned account.

MatthewBozin:
Thanks @entrotech! Would you like me to open an issue to look into the stakeholder search query?
Also, I'd be happy to work on the AWS login, what would be the best place to discuss this further?

PR #1592, opened 2023-03-14
Refactored tenant parameters

junjun107:


client/src/helpers/Configuration.js
```diff
@@ -1,57 +1,10 @@
...
-      };
-    default:
-      return {
-        center: { latitude: 34.0354899, longitude: -118.2439235 },
-        zoom: 11,
-      };
-  }
-})();
+export const tenantDetails = TENANT_CONFIG[tenantId];
+console.log(tenantId, TENANT_CONFIG);
```
hanapotski, line 9: we can remove the log 👌

junjun107: Removed

Issue #1594, opened 2023-03-15
Bug: Deployment failing


entrotech:
### Overview
Starting with PR #1589, the Docker build is failing with the message shown in the Resources/Instructions section.

### Action Items
- [ ] Diagnose and Fix

### Resources/Instructions
GitHub action message:
```Step 10/44 : RUN npm run build
 ---> Running in 066fbaaf4928

> foodoasis-client@1.0.70 build
> craco build

Creating an optimized production build...
Failed to compile.

Error: [BABEL] /app/src/index.js: React Refresh Babel transform should only be enabled in development environment. Instead, the environment is: "production". If you want to override this check, pass {skipEnvCheck: true} as plugin options. (While processing: "/app/node_modules/react-scripts/node_modules/react-refresh/babel.js")
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
    at Generator.next (<anonymous>)
    at cachedFunction.next (<anonymous>)
    at loadPluginDescriptor.next (<anonymous>)
    at loadPluginDescriptors.next (<anonymous>)
    at Generator.next (<anonymous>)
    at loadFullConfig.next (<anonymous>)
    at transform.next (<anonymous>)


The command '/bin/sh -c npm run build' returned a non-zero code: 1
Error: exit status 1```

**Comment by entrotech** [comment]
Tried changing to NODE_ENV = "production" in the Dockerfile, but this caused a different error:
```
Step 10/44 : RUN npm run build
 ---> Running in 2a30[94](https://github.com/hackforla/food-oasis/actions/runs/4431772536/jobs/7778333201#step:4:95)15a880

> foodoasis-client@1.0.70 build
> craco build

node:internal/modules/cjs/loader:[107](https://github.com/hackforla/food-oasis/actions/runs/4431772536/jobs/7778333201#step:4:108)5
  const err = new Error(message);
              ^

Error: Cannot find module 'react-scripts/package.json'
Require stack:
- /app/node_modules/@craco/craco/dist/lib/cra.js
- /app/node_modules/@craco/craco/dist/scripts/build.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1075:15)
    at Function.resolve (node:internal/modules/cjs/helpers:[116](https://github.com/hackforla/food-oasis/actions/runs/4431772536/jobs/7778333201#step:4:117):19)
[...]

PR #1597, opened 2023-03-16
1594 react refresh deployment fail

entrotech:
Take 2 on fixing #1594


entrotech:
Hannah finished fixing this problem.

PR #1622, opened 2023-04-01
1572 Add strong password validation for Register & Reset Password page

### LINKED ISSUE #1572: Add strong password validation on signup/change password, number, uppercase, lowercase.
### Overview
Add strong password validation on signup/change password, number, uppercase, lowercase.

devnikesh:
### Overview
Added strong password validation on the signup and Reset Password pages, requiring passwords to contain at least one number, one uppercase letter, one lowercase letter, and one special character.  With this new validation in place, users will be prompted to create stronger passwords that will better protect their personal information and accounts.

Fixes #1572

### DIFF
--- a/client/src/components/Account/ResetPassword.js
+++ b/client/src/components/Account/ResetPassword.js
@@ -38,7 +38,10 @@ const styles = (theme) => ({
 const validationSchema = Yup.object().shape({
   token: Yup.string().required("Token is required"),
   password: Yup.string()
-    .min(8, "Password must be 8 characters at minimum")
+    .matches(
+      /^.*(?=.{8,})((?=.*[!@#$%^&*()\-_=+{};:,<.>]){1})(?=.*\d)((?=.*[a-z]){1})((?=.*[A-Z]){1}).*$/,
+      "Password must contain at least 8 characters, one uppercase, one number and one special case character"
+    )
     .required("Password is required"),
   passwordConfirm: Yup.string()
     .required("Confirm your password")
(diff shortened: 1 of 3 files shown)


client/src/components/Account/ResetPassword.js
```diff
@@ -38,7 +38,10 @@ const styles = (theme) => ({
 const validationSchema = Yup.object().shape({
   token: Yup.string().required("Token is required"),
   password: Yup.string()
-    .min(8, "Password must be 8 characters at minimum")
+    .matches(
+      /^.*(?=.{8,})((?=.*[!@#$%^&*()\-_=+{};:,<.>]){1})(?=.*\d)((?=.*[a-z]){1})((?=.*[A-Z]){1}).*$/,
```
hanapotski, line 42: can we move the regex on a helper file so we can just reuse it?😅

client/src/components/Account/ResetPassword.js
```diff
@@ -38,7 +38,10 @@ const styles = (theme) => ({
...
+      "Password must contain at least 8 characters, one uppercase, one number and one special case character"
```
hanapotski, line 43: Same with this error message 👌

client/src/components/Account/ResetPassword.js
devnikesh: Stored it in the `Constants.js` file

client/src/components/Account/ResetPassword.js
devnikesh: I see there are similar validation message used in more than one component, should I also store them. For e.g: "Password is required", "Email is required"

devnikesh: I see there are similar validation message used in more than one component, should I also store them? 🤔  For e.g: "Password is required", "Email is required"

hanapotski (approved):

PR #1713, opened 2023-06-09
Found Bug in Links color

### LINKED ISSUE #1709: Bug: Normal Links are not the correct color.
### Overview
Normal Links are being rendered with a color of #336699, instead of the desired color of #1976D2, even though the MuiLink override in Link.js specified the correct color for the &:link pseudoclass. This appears to be because Link is built on top of Typography component, which has it color specified in the Typography.js file.

### Action Items
- [x] Figure out how to make the normal link color correct in the theme.

### Resources/Instructions
image

93Belen:
Fixes #1709 

The bug was located in `theme/overrides/Link.js`. The variants applied to the Links were the ones intended for the Typography.

The MUI docs for Link say that 

> inherit Applies the theme typography styles.
https://mui.com/material-ui/api/link/

I am not familiar with the Overrides in MUI but by changing the overrides and adding the desired styles in a new variant, the color was corrected.
I also added the variant to a few links. We might need to make another Issue to find all the Links that need to have the new "primary" variant added.

<details><summary>Link color before my changes</summary

before changes color

</details>

<details><summary>Link color after my changes</summary>

after changes color

</details>

### DIFF
--- a/client/src/theme/overrides/Link.js
+++ b/client/src/theme/overrides/Link.js
@@ -9,6 +9,9 @@
 export default function Link(theme) {
   return {
     MuiLink: {
+      defaultProps: {
+        default: "primary"
+      },
       variants: [
         {
           props: { variant: "icon" },
--- a/client/src/theme/overrides/Link.js
+++ b/client/src/theme/overrides/Link.js
@@ -16,26 +16,30 @@ export default function Link(theme) {
             textDecoration: "none",
           },
         },
-      ],
-      styleOverrides: {
-        inherit: {
+        {
+          props: { variant: "primary" },
+          style: {
           textDecoration: "none",
           padding: "2px 1px 0",
           borderBottom: "1px solid",
           "&:link": {
             color: theme.palette.link.normal,
           },
-          "&:visited": {
-            color: theme.palette.link.visited,
+          "&:a": {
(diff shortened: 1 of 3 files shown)


client/src/theme/overrides/Link.js
```diff
@@ -9,22 +9,28 @@
 export default function Link(theme) {
   return {
     MuiLink: {
+      defaultProps: {
+        default: "primary"
```
hanapotski, line 13: @93Belen just noticed this. It should be `variant: "primary"` 😅

client/src/theme/overrides/Link.js
```diff
@@ -9,22 +9,28 @@
...
         },
-      ],
-      styleOverrides: {
-        inherit: {
+        {
+          props: { variant: "primary" },
+          style: {
           textDecoration: "none",
           padding: "2px 1px 0",
           borderBottom: "1px solid",
```
hanapotski, line 27: also think we need `color: theme.palette.link.normal` here 🤔

client/src/theme/overrides/Link.js
93Belen: https://github.com/hackforla/food-oasis/pull/1720

PR #1717, opened 2023-06-11
1471 disable button when schema invalid

VirginiaWu11:


VirginiaWu11:
instead of disabling the save button when there is an error, scrolling to the field with the error is more user friendly.

VirginiaWu11:
fixed with PR #1716

PR #1804, opened 2023-09-17
Update Styling System to Use xs Prop in Material-UI Components in CorrectionDialog.js

sisi-in-tech:
- Resolved the styling for the CorrectionDialog.js to align with Material-UI version 5. 

- CorrectionDialog.js before image 
before-image

-  CorrectionDialog.js after image 
after-image


kurikurichan (approved):
LGTM

junjun107 (commented):
I don't think we need the sx styling here 🧐. Textarea is a custom ui component in the codebase and it already has styling. I vote for just remove the makestyle and classes.

sisi-in-tech:
After testing, it was found that removing the sx prop from Textarea in CorrectionDialog.js has the same effect as before when it had the sx prop. This section has been updated. 

1. Before image : 
origial_image

2. After image :
after_delete_sx

junjun107 (approved):
Very clean. Look good.

PR #1896, opened 2023-10-29
1691 fix search autocomplete disappear bug

### LINKED ISSUE #1691: Bug: Address autocomplete sometimes disappears even when there are matches (bug?)
### Overview

Bug discovered while doing #1668 that prevents people from entering or resolving addresses when typing them in to the search field on both the homepage and the map screen..

When entering a text address into the address field, the autocomplete addresses pop-up that appears will often disappear and reappear.

### How to replicate

To replicate:
1. Go to foodoasis.la landing page
2. Enter 90045, one character at a time.  The autocomplete pop-up shows matching addresses as each character is typed
3. Seemingly randomly, at the 4th or 5th digit for me, the autocomplete options pop-up disappears.  Sometimes it reappears when adding additional characters.

Seems like this is unintentional and not working as designed.

VirginiaWu11:
- closes #1691 

https://github.com/hackforla/food-oasis/assets/79432798/abd63789-1c8f-47e4-92a3-062686742fb0

### DIFF
--- a/client/src/components/FoodSeeker/AddressDropDown.js
+++ b/client/src/components/FoodSeeker/AddressDropDown.js
@@ -26,7 +26,6 @@ export default function AddressDropDown({ showSearchIcon }) {
     const safeValue = typeof delta === "string" ? delta : delta.target.value;
     setInputVal(safeValue);
     if (safeValue) {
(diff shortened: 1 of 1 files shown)


junjun107 (approved):
Nice

PR #1941, opened 2023-12-13
removes redundancy for preview component in mobile view

qiqicodes:
`StakeholderPreview` component currently is in both `ResultList` component and Mobile Layout.  while `ResultList` component is also in Mobile layout .

The app behaves the same when removing the preview component from the Mobile Layout. 

https://github.com/hackforla/food-oasis/assets/68657634/a8f79674-8260-41f8-90b8-7b4556ec09a7


qiqicodes:
Currently the fix works for both desktop and mobile on computer's chrome browser. But when testing on the mobile device's chrome browser(device uses iOS version 16.6.1), the filter panel doesn't show up, and the  `More Filter` button toggles the list state.

junjun107 (approved):
LGTM

PR #1950, opened 2023-12-24
Fixes server format script

qiqicodes:
Updates the server format script
- format all typescript and javascript files within the server dir with the existing prettier setting
- ran script to format files


kurikurichan (approved):
LGTM

PR #1981, opened 2024-01-24
Update color button stakeholderdetails

### LINKED ISSUE #1965: Update color of  buttons in StakeholderDetails
### Overview
For now, we are focusing on changing just he color of the button, border, and text within (switch from old primary blue to light gray as showing in Figma)

old color image
new color image

old color image
new color  image

image
image

### Action Items
Use the following Figma links to see the expected color needed
https://www.figma.com/file/9hPmNSsyaH2VMGNz5mffmI/FOLA-Design-%232?type=design&node-id=640-125&mode=design&t=mkQ8pAJGRjTkm6PV-0

93Belen:
Fixes #1965 

**Since the Issue only required to change colors, I didn't change anything else in the style**

- I added a `gay variant` to the Button override file.
- I added this variant to the buttons needed
- I changed the color of the button Back to List

_I notice that the Figma files don't show the styles for hover effect or disabled state, I added the hover effect but the gray variant still need the disabled style._

Hover style added: 
Screenshot 2024-01-24 at 6 26 56 AM


client/src/theme/overrides/Button.js
```diff
@@ -105,6 +105,28 @@ export default function Button(theme) {
...
+          backgroundColor: theme.palette.common.white,
+          borderStyle: "solid",
+          borderWidth: "1px",
+          borderColor: '#999999',
+          fontSize: ".8125rem",
+          "@media (min-width:600px)": {
+            fontSize: "0.8125rem",
+          },
+          "&:hover": {
+            backgroundColor: '#E1E1E1',
```
qiqicodes, line 120: The background color should be "#E2E5EE"  instead of "#E1E1E1" according the Figma.

Screenshot 2024-01-26 at 4 35 33 PM

client/src/theme/overrides/Button.js
```diff
@@ -105,6 +105,28 @@ export default function Button(theme) {
...
+          },
+          "&:active": {
```
qiqicodes, line 122: The active state is not listed in Figma. @JohnHaoHuang, can you clarify the active state?

qiqicodes (changes requested):

client/src/theme/overrides/Button.js
93Belen: Omg thats so helpful, I didn't see that in the sigma and I made up the hover state! I wasn't to far from the real one haha

93Belen:
@qiqicodes Hey! I added the two changes you suggested :) Regarding the Active State, I changed the background color to the same one as for the hover state, cause that seems to be the pattern in the other styles. I think we can leave it as placeholder until they tell us otherwise.
I hope everything looks ok now, thanks!

PR #2043, opened 2024-03-19
implement feature flag

### LINKED ISSUE #2042: Implement feature flag on client
The goal of this ticket is to determine whether the current user is authorized to access a particular feature. It replaces the old environment variable feature flag system. 

Backend:

- [x] Update getByEmail in account-service to retrieve both user details and their feature names list.
- [x] Update features types

Frontend: 

- [x] Modified useFeatureFlag hook to use userContext to check if current user has access to the specific feature

junjun107:
This replace the old feature flag using a variable 
close #2042

### DIFF
--- a/client/src/components/Admin/Features.js
+++ b/client/src/components/Admin/Features.js
@@ -45,15 +45,6 @@ function createData(featureName, users, featureId) {
     })),
   };
 }
-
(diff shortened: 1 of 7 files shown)


qiqicodes (approved):
lgtm

PR #2062, opened 2024-03-27
fixes query

### LINKED ISSUE #2061: Bug: Query for Listing Fetches Listings from All Tenants.
#### Describe the bug 

The application fetches all listings from all tenants, when it should only get listings for the current tenant. E.g for la.foodoasis.net, the application should only download listings for tenantId = 1.  This violates the segregation of listings by tenantId, and also significantly impacts performance, since many listings are retrieved that aren't needed.

(if applicable, please specify platform (iOS, Android, Windows, Mac version) and browser)

1. Go to map page.  You can look at Developer console Network tab to see that too many listings are downloaded or scroll the map to McKinney TX, or Oahu and see that listings from other regions ar appearing on the map and list.

Only listings for the current tenant (i.e., region) should be retrieved.

(if you have a suggested solution, you can add it here)
Add tenantId to the stakeholderbest search criteria.

-

- Screenshot (drag in)
- Video recording (drag in)

#### Device configuration
- Device: [e.g. type of smartphone, tablet, desktop computer]
- OS version: [e.g. iOS, Android version #]
- Browser [e.g. Chrome, Firefox, Safari, etc.]
- Browser version [e.g. 22]

### LINKED ISSUE #2067: Bug: Incorrectly showing pins from outside the user's current region
#### Describe the bug 

Severity: 2/5 (where 5 is the worst)

On the old site, we filtered pins down to the user's current region — so for example, foodoasis.la/la.foodoasis.net showed only LA-region locations.  McKinney.foodoasis.la only showed McKinney locations.  

With the recent Jan 2024 release, we are now showing all pins worldwide to everyone. 

This isn't as expected — currently, we want to limit the pins we show because certain organizations are responsible for certain pins, and showing pins outside of a site's responsibility could lead to problems such as incorrect information being shared.

This is particularly important when we're doing data sharing / giving access to our database to Hack for LA data miners (from other projects).  Data from regions that are closed or are very outdated may be showing up in the data feeds they pull down.  This needs to be made clear to data miners.

1. Go to any location on the map on la.foodoasis.net
2. Zoom out to a world-wide view

You will see pins outside of the LA region, such as in Hawaii (a region's site that never launched) and McKinney, Texas.

- Locations only from the user's region, based on the website they're visiting.

- Showing pins from all regions at once, including ones from deprecated/disabled sites, such as Santa Rosa.

#### To fix:

Re-implement region filtering for both map's pins and listings as on previous site

Optional: Discuss with team leads if we should have a way to mark regions as 'off' so that data miners don't accidentally pull down outdated data from those inactive/test regions (such as Santa Rosa/northern California and Hawaii)

Screenshot 2024-03-28 at 1 26 15 PM

#### Device configuration
- Device: Mac
- OS version: Sonoma
- Browser: Safari

qiqicodes:
closes #2061
closes #2067

### DIFF
--- a/server/app/services/stakeholder-best-service.ts
+++ b/server/app/services/stakeholder-best-service.ts
@@ -58,8 +58,8 @@ const selectAll = async ({ tenantId }: { tenantId: string }) => {
   s.allow_walkins, s.hours_notes, s.tags
   FROM stakeholder_best s
   LEFT JOIN neighborhood n on s.neighborhood_id = n.id
-  WHERE 1 = ANY(s.category_ids)
-  OR 9 = ANY(s.category_ids)
+  WHERE (1 = ANY(s.category_ids)
+  OR 9 = ANY(s.category_ids))
   AND s.tenant_id = ${tenantId}
   `;
 


hanapotski:
were you able to check if this really fixes the issues mentioned above?🤔 I'm not sure why the category ids would have anything to do with it 🤔

qiqicodes:
> were you able to check if this really fixes the issues mentioned above?🤔 I'm not sure why the category ids would have anything to do with it 🤔

It's the `or` logic, I have double checked with the query from the database.

hanapotski:
gotcha 👌

hanapotski (approved):

PR #2073, opened 2024-04-01
Fix type error

hanapotski:


junjun107 (approved):
LGTM 
Thanks for fixing it.

Issue #2097, opened 2024-04-05
Refactor frontend routes


hanapotski:
This should just be a refactor and shouldn't affect any functionality.

PR #2082, opened 2024-04-02
fixes meal pantries button toggle bug

qiqicodes:
fixes #2078 

fixed desktop view

https://github.com/hackforla/food-oasis/assets/68657634/6a6414fc-6f14-4aec-9935-cf99223590a9

fixed mobile view

https://github.com/hackforla/food-oasis/assets/68657634/8019fdbc-866d-4f89-8b46-41c7ec8630f3

PR #2098, opened 2024-04-05
Refactor frontend routes

hanapotski:
Addresses #2097


client/src/Routes.js
```diff
@@ -0,0 +1,206 @@
...
+  );
+}
+
+function AdminWrapper() {
+  return (
+    <>
+      <Outlet />
+    </>
+  );
+}
```
hanapotski, line 206: This isn't necessary but we used to wrap some admin routes with a div with styles and if we ever decide to add any styles specific to all admin routes, it should go here

junjun107:
I don't see the map.

qiqicodes:
ResultsMap currently only renders on Mobile view, and not on Tablet and Desktop Views.

Nit: The advance filter feature's `Pantry` and `Meal` toggle button currently on works on Mobile View, and not on Tablet and Desktop Views. Rebase #2082 to implemented the change.

client/src/Routes.js
```diff
@@ -0,0 +1,206 @@
...
+              path="organizationimport"
+              element={
+                <PrivateRoute roles={["isAdmin"]}>
+                  <ImportFile />
+                </PrivateRoute>
+              }
+            />
+          </Route>
+
+          <Route exact path="fallback" element={<Fallback />} />
```
qiqicodes, line 184: To handle undefined admin paths, add a catch-all route below `organizationimport` route, redirecting to the fallback route:

```    
// Below `organizationimport` route, add:
   <Route path="*" element={<Navigate to="/fallback" replace />} />
</Route> // Closes admin routes
// Place the fallback route outside the admin routes
   <Route exact path="fallback" element={<Fallback />} />
```

qiqicodes (changes requested):
Screen Shot 2024-04-08 at 4 21 54 PM

Bug: On the login page, if incorrect credentials are entered, the expected toast message "The password is incorrect, please check it and try again or use the Forgot Password feature." does not appear.

Bug: `Verify Email` Button from the email doesn't direct me to the verification confirmation page for me locally.

client/src/Routes.js
```diff
@@ -0,0 +1,227 @@
...
+
+          <Route path="organizations" element={<SearchResults />} />
+          <Route path="suggestion" element={<Suggestion />} />
+          <Route path="donate" element={<Donate />} />
+          <Route path="about" element={<About />} />
+          <Route exact path="faqs" element={<Faq />} />
+          <Route exact path="contact" element={<Contact />} />
+
+          {/* Admin routes */}
+          <Route path="/admin" element={<AdminWrapper />}>
```
qiqicodes, line 90: Nit: can remove the "/" from path
```
<Route path="admin" element={<AdminWrapper />}>
```

qiqicodes (approved):
lgtm

PR #2106, opened 2024-04-09
Fix date bug

### LINKED ISSUE #2104: Bug - Page crashes on scroll due to 'e.getDate is not a function' error in event handler
image

### Steps to reproduce:

1. Go to stakeholder lists
2. Place cursor at one of the stakeholder 
3. Use mouse center wheel to scroll down very fast
4. App crash, date.getDate is not a function error in console

hanapotski:
Fixes #2104


andyarensman (approved):
LGTM

qiqicodes (changes requested):
This resolves the TypeError, but it leads to a duplicate key error in the accordion. Could you also address that issue? Incorporating the following adjustment at line 384 in the `StakeholderPreview.js` file should resolve it:

```
key={JSON.stringify({ ...hour, index })}
```

hanapotski:
@qiqicodes might be unrelated 🤔 I couldn't replicate it but I updated the key with the index 👌

PR #2107, opened 2024-04-09
Fix back button bug

### LINKED ISSUE #2092: Serious Bug: Using browser's 'back' from org detail loses place in the list
Priority: 4/5 (5 is most urgent) - This is very important as this creates big friction and affects almost all food-seekers actively reviewing listings.

#### Describe the bug 

When viewing an org's details, then clicking **on the browser 'back' button**, my place is not held in the scrolled list as it restarts the list back from the top and forces me to rescroll down to pick up where I left off. 

This makes it very difficult to browse and penalizes the user who gets more information about an organization.

The goal of a user is to find an organization… typically they start at a long list, scroll down to something that looks interesting, get more details, exit back to the list and continue scrolling the list from where they left off.  But our site now prevents this.

1. Do a search and view a list of organizations
2. Scroll down list to the 10th organization in the list, which is off-screen at first
3. Click on "Details"
4. View the organization's details.  Use browser's "Back" button
5. View the list of organizations, but unfortunately, list has scrolled back to the top item, and now it's a pain to try finding the 11th item in the list to continue your research

The list should be exactly as it was when the user last interacted with it, so if the 10th item was in the middle of the screen, then it should again be there.

I would also suggest: consider using an animated yellow-beige highlight of the org's tile in the list for a few seconds of org's tile when returning from it, for example, will also make it easier to track that this is the org page that you've just returned from.

The list as scrolled to the beginning, which means I've lost my place.  This also makes it very difficult to browse org details.

#### Additional details / screenshot / video

https://github.com/hackforla/food-oasis/assets/3376957/c34aef54-61e5-4bd9-8b8a-24c502def912

#### Device configuration
- Device: Desktop (and probably mobile, too)
- OS version: Mac
- Browser Safari

hanapotski:
Fixes #2092

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.js
+++ b/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.js
@@ -204,6 +204,16 @@ const StakeholderDetails = ({ onBackClick, isDesktop }) => {
     isAlmostOpenFlag &&
     calculateMinutesToOpening(stakeholderHours, tenantTimeZone);
 
+  useEffect(() => {
+    window.addEventListener(
+      "popstate",
+      function () {
+        onBackClick();
+      },
+      false
+    );
+  }, [onBackClick]);
+
   return (
     <>
       <SEO


junjun107 (approved):
LGTM 
Thank you for showing how windows popstate  event handler works  🥳

qiqicodes (approved):
lgtm

PR #2123, opened 2024-04-16
Fix null bug

### LINKED ISSUE #2120: Bug: Error loading the application when organization does not have a phone number
When a user is on the organization detail view, and enters an address and clicks return on the search bar on the food seeker view, the application will be unusable. 

This is the error log
Screen Shot 2024-04-15 at 4 40 45 PM

hanapotski:
Fixes #2120


qiqicodes (changes requested):
The fix looks great! 

One small note: please run `npm run format` to apply Prettier formatting.

qiqicodes (approved):

PR #2142, opened 2024-05-21
2045 remove redundant tooltip

### LINKED ISSUE #2045: Remove the "Use my current location" tooltip
The tooltip has the same copy as the the button text so we just want to remove it.

andyarensman:
Closes #2045 

Removed the redundant "use my current location" tooltip, but kept the tooltip when your location is disabled:

chrome_JikjW8F4ou

chrome_bPLmgwyMPK

### DIFF
--- a/client/src/components/FoodSeeker/Home.js
+++ b/client/src/components/FoodSeeker/Home.js
@@ -73,9 +73,9 @@ const Home = () => {
   };
 
   useEffect(() => {
-    const loadingSpinner = document.getElementById('home-loading-spinner');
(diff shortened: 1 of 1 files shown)


93Belen (changes requested):
It works on Desktop, but doesn't show the tooltip alert when testing on mobile. I guess it only works with Hover or Click but not with Touch? I'm not sure if this is something you want to address in this Issue or it's technically a different and pre-existing problem.

andyarensman:
@93Belen Good find! I found a quick fix

93Belen (approved):
LGTM! :)

PR #2147, opened 2024-05-31
Fixes center on current location bug

### LINKED ISSUE #2065: Serious Bug: GPS button doesn’t re-center on current location
#### Describe the bug 

Severity: 5/5 (where 5 is worst)

Using the “GPS” button in the map view correctly gets the user’s location and correctly places a marker, but it doesn’t update the map to match, so the user’s marked location is far off-screen.

This is unexpected behavior from the user, and I think a regression.

1. Go to any location on the map far from where you are located
2. Click on GPS button (upper right corner, next to the address field)
3. Error > your marked location on the map is not centered, and may be completely off-screen, especially if you are zoomed in.

- Map should scroll or jump to your current location and marker on the map (assuming we were able to get a location for you)

A smooth scroll might be a nice bonus, but is not necessary.

- Marker is indeed on map, but often off-screen.

#### Device configuration
- Device: Desktop 
- OS version: Mac
- Browser: Safari

qiqicodes:
fixes #2065

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.js
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.js
@@ -98,7 +98,18 @@ const ResultsMap = (
     await loadMarkerIcons(map);
     setMarkersLoaded(true);
     setInteractiveLayerIds([MARKERS_LAYER_ID]);
-  }, []);
+    mapRef.current?.flyTo({
+      center: [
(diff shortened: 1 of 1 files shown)


entrotech:
I merged this to give it a try, but I don't think it is working.  I'm in San Diego now, and when I try the "GPS" button, it places the map centered on Moreno Valley, and San Diego is far enough south of there, that it is off the map. Since it is merged, you should create a new PR.

PR #2174, opened 2024-07-25
Fix filterpanel bug and flyto clean up

### LINKED ISSUE #2168: Text panel obscures map and pin on mobile, off-center on desktop
severity 5/5
Near impossible to see location on map on mobile devices
Maps are centered as if text panel doesn’t exist leading to obscured or weird alignments. Map should center using its visible area, not the entire screen. 

### Problem:
On mobile, tapping a result to get details loads org detail view, but layout of split screen obscures map and pin, and pin is off-screen. 

Desktop has a similar problem.

To reproduce the desktop version of this bug:
1. From landing page, enter an address. On desktop, map Pin appears in the correct place at first.
IMG_1253

2. Do a search again from the map page and the map now re-centers in the wrong place.
IMG_1252

### User goal: 
to get see the cross streets of the location indicated on the map so i tap on the icon

### my expectation: 
To zoom in to that specific location and see more information of the name and hours of operation (maybe covering maximum bottom half, and leave the upper half of the screen on the map  showing the selected location zoomed in.

### actual result:
80% screen covered by TEXT on location's information etc. Covering the map.
So I swipe down to open the window to map again.
BUT the map window has migrated to a different location, no where close to the selected location.

### screenshots and video 

https://github.com/hackforla/food-oasis/assets/3376957/87062398-06a2-4bcd-99bf-9ecd2cce789f

https://github.com/user-attachments/assets/be6ff243-2863-43cc-b2f1-d940acb4e386

image

### possible solutions

make map larger - see figma designs - I believe it’s supposed to be 50% text at first? — Bryan

Ensure that with new split-screen view that pin is centered in map viewport at first load.  This might mean calculating visible map and centering it as text panel is resized.

for example if text is 80% of screen, map is remaining 20% and pin should be at approx 10%.

Ideally the selected pin and the user’s location are both visible on the map at the same time.  This will give them a sense of location relative to where they are now.

Stretch goal: if pin is centered, then keep it centered as user resizes text panel. 

### related issues
-  #2028

hanapotski:
Closes #2168 

- fix filter panel bug on mobile
- remove forward ref and extra mapRef
- clean up flyTo

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -10,6 +10,7 @@ import * as analytics from "../src/services/analytics";
 import { AppStateProvider } from "./appReducer";
 import SEO from "./components/SEO";
 import AppRoutes from "./Routes";
+import { MapProvider } from "react-map-gl";
 
 function App() {
   useEffect(() => {
@@ -29,9 +30,11 @@ function App() {
                   url={window.origin}
                   description="Food Oasis is a non-profit, volunteer-run directory of free food resources in the Los Angeles area."
                 />
-                <Router>
-                  <AppRoutes />
-                </Router>
+                <MapProvider>
+                  <Router>
+                    <AppRoutes />
+                  </Router>
+                </MapProvider>
               </ThemeProvider>
             </UserProvider>
           </ToasterProvider>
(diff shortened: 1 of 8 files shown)


SAUMILDHANKAR (approved):
Thanks a lot Hannah for fixing this issue. Looks good to me. As discussed on Saturday, just noting down my observation: when searching for a location in the search bar on mobile and pressing enter, I see the location icon not at exactly the center but just a little off, it is very minor. Thank you.

Issue #2160, opened 2024-06-19
Add ability to be able to toggle feature flag for everyone 


junjun107:
### Overview

Create a toggle to turn on feature flag for everyone

PR #2181, opened 2024-08-09
implement global enable feature functionality

junjun107:
#2160 

Global Enable toggle button is added in Admin Feature Flag table. When toggle button is green, feature is globally enabled for everyone. 

Screenshot 2024-08-07 at 9 08 39 PM

Demo: 
Step 1: Enable feature globally
Step 2: Make sure account is not logged in
Step 3: Advanced Filter button group (shown in green) is visible. 
not Logged in

### DIFF
--- a/client/src/components/Admin/Features.js
+++ b/client/src/components/Admin/Features.js
@@ -17,6 +17,8 @@ import {
   TableRow,
   TextField,
   Typography,
+  FormControlLabel,
+  Switch,
 } from "@mui/material";
 import {
   Delete as DeleteIcon,
@@ -28,6 +30,7 @@ import { useFormik } from "formik";
 import React, { useEffect, useState } from "react";
 import * as Yup from "yup";
 import { useFeatureToLogin } from "../../hooks/useFeatureToLogin";
+import { useFeatures } from "../../hooks/useFeatures";
 import * as accountService from "../../services/account-service";
 import * as featureService from "../../services/feature-service";
 import * as featureToLoginService from "../../services/feature-to-login-service";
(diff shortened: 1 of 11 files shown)


hanapotski:
this might or might not be blocking 😅 I think the first time the page loads, we don't have the feature flag data so we see this flash. We might just need a new ticket to address it. We need to get the feature flag data before we render anything 🤔 

https://github.com/user-attachments/assets/9540e64f-02c8-4430-8a4c-58b54b03431c

client/src/components/Admin/Features.js
```diff
@@ -48,21 +51,39 @@ const Features = () => {
...
-      featureToLoginId: feature.ftl_id,
-      history: feature.users.map((user) => ({
-        loginId: user.login_id,
-        firstName: user.first_name,
-        lastName: user.last_name,
-        email: user.email,
-        featureToLoginId: user.featureToLoginId || feature.ftl_id,
-      })),
-    }));
+    if (featureToLoginLoading || featuresLoading) {
```
hanapotski, line 57: we could just check for `featuresData` before we use it. Returning something from a useEffect isn't necessary unless we need to do some clean up 🤔

client/src/hooks/useFeatureFlag.js
```diff
@@ -1,10 +1,17 @@
...
-  if (!user || !user.features) {
-    return false;
+  const { data: featureFlags } = useFeatures();
+  const featureFlag = featureFlags.find((feature) => feature.name === flagName);
+
+  if (featureFlag && featureFlag.is_enabled) {
+    return true;
+  }
+  if (user && user.features && user.features.includes(flagName)) {
+    return true;
```
hanapotski, line 13: I think we can move these conditions to a variable so it's easier to read and follow like
```
const isFeatureEnabled = ...
const userHasFeature = ...
```
so we can just check `if (isFeatureEnabled || userHasFeature)`

client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.js
```diff
@@ -182,7 +182,7 @@ const ResultsMap = (
 
   const listPanelLeftPostion = isListPanelOpen ? 524 : 0;
   const filterPanelLeftPostion = isFilterPanelOpen ? 340 : 0;
-
+  console.log(hasAdvancedFilterFeatureFlag);
```
hanapotski, line 185: we can remove this log 👀

client/src/components/Admin/Features.js
```diff
@@ -48,21 +51,30 @@ const Features = () => {
...
+          ? featureToLogin.users.map((user) => ({
+              loginId: user.login_id,
+              firstName: user.first_name,
+              lastName: user.last_name,
+              email: user.email,
+              featureToLoginId: user.featureToLoginId || featureToLogin.ftl_id,
+            }))
+          : [],
+      };
+    });
```
hanapotski, line 75: can we wrap this in a check for `featureToLoginData && featuresData`?

hanapotski (approved):

PR #2183, opened 2024-08-10
Add client side validation for pinterest

### LINKED ISSUE #2182: Add client-side validation and tooltip for Pinterest contact detail field
### Overview
We would like the data in our contact field to be accurate. In this issue, we will add client-side validation for the Pinterest contact field.

### Action Items
- [x] Add client-side validation which prompts user in case of an error. 
- [x] Test it in your dev environment.

### Resources/Instructions
- Related file: https://github.com/hackforla/food-oasis/blob/develop/client/src/components/Admin/OrganizationEdit/ContactDetails.js

monicakrystal:
Closes #2182 

Before
Screenshot 2024-08-10 at 2 24 45 PM

After ! 
Screenshot 2024-08-10 at 2 25 03 PM

### DIFF
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -53,10 +53,19 @@ const validationSchema = Yup.object().shape({
   hours: Yup.array().of(HourSchema),
   twitter: Yup.string()
     .matches(
-      /^https?:\/\/(www\.)?(twitter\.com|x\.com)\/.*/,
-      "Invalid URL, e.g. 'https://twitter.com/ or https://x.com/'"
+      /^@([a-zA-Z][a-zA-Z0-9_]{0,14}|[0-9][a-zA-Z0-9_]{0,14})$/,
+      "Valid Twitter username required."
     )
-    .required("Full Twitter/X URL is required."),
+    .required("Valid Twitter username is required."),
+
+  pinterest: Yup.string()
+    .matches(
+      /^[a-zA-Z0-9.]{3,15}$/,
+      "Valid Pinterest username is required and cannot include spaces."
+    )
+    .required("Valid Pinterest username is required.")
+    .notOneOf([""], "Username cannot be empty or just spaces"),
+
   selectedCategoryIds: Yup.array().min(
     1,
     "You must select at least one category"
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -62,7 +62,7 @@ const validationSchema = Yup.object().shape({
     )
     .matches(
       /^@?(?=.*[a-zA-Z])[a-zA-Z0-9_]{3,30}$/,
-      "Valid Pinterest username is required. It must be 3 to 30 characters long, contain only letters, numbers, and underscores, must include at least one letter, and cannot contain spaces or punctuation marks. It may optionally start with an '@' symbol."
+      "Valid Pinterest username is required."
     )
     .nullable(),
 
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -61,8 +61,8 @@ const validationSchema = Yup.object().shape({
       originalValue.trim() === "" ? null : value
     )
     .matches(
-      /^[a-zA-Z0-9.]{3,15}$/,
-      "Valid Pinterest username is required and cannot include spaces."
+      /^@?(?=.*[a-zA-Z])[a-zA-Z0-9_]{3,30}$/,
+      "Valid Pinterest username is required. It must be 3 to 30 characters long, contain only letters, numbers, and underscores, must include at least one letter, and cannot contain spaces or punctuation marks. It may optionally start with an '@' symbol."
     )
     .nullable(),
 
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -53,10 +53,7 @@ const validationSchema = Yup.object().shape({
   hours: Yup.array().of(HourSchema),
   twitter: Yup.string()
     .transform((value, originalValue) => (originalValue === "" ? null : value))
-    .matches(
-      /^@([a-zA-Z][a-zA-Z0-9_]{0,14}|[0-9][a-zA-Z0-9_]{0,14})$/,
-      "Valid Twitter username required."
-    )
+    .matches(/^(?:@?[a-zA-Z0-9_]{1,15})$/, "Valid Twitter username required.")
     .nullable(),
 
   pinterest: Yup.string()
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -52,19 +52,22 @@ const validationSchema = Yup.object().shape({
   email: Yup.string().email("Invalid email address format"),
   hours: Yup.array().of(HourSchema),
   twitter: Yup.string()
+    .transform((value, originalValue) => (originalValue === "" ? null : value))
     .matches(
       /^@([a-zA-Z][a-zA-Z0-9_]{0,14}|[0-9][a-zA-Z0-9_]{0,14})$/,
       "Valid Twitter username required."
     )
-    .required("Valid Twitter username is required."),
+    .nullable(),
 
   pinterest: Yup.string()
+    .transform((value, originalValue) =>
+      originalValue.trim() === "" ? null : value
+    )
     .matches(
       /^[a-zA-Z0-9.]{3,15}$/,
       "Valid Pinterest username is required and cannot include spaces."
     )
-    .required("Valid Pinterest username is required.")
-    .notOneOf([""], "Username cannot be empty or just spaces"),
+    .nullable(),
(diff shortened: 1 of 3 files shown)


client/src/components/Admin/OrganizationEdit/ContactDetails.js
```diff
@@ -79,7 +79,11 @@ export default function ContactDetails({
         </Grid>
         <Grid item sm={6} xs={12}>
           <div>
-            <Label id="pinterest" label="Pinterest" />
+            <Label
+              id="pinterest"
+              label="Pinterest"
+              tooltipTitle="URL must start with 'https://pinterest.com/'"
```
hanapotski, line 85: in the last stand up meeting, we've talked about adding a clickable link for the user so they can click it and test that the link works. Maybe we can add it next to the tooltip. It can just be a `<Button variant="text" size="small">` 👌

client/src/components/Admin/OrganizationEdit.js
```diff
@@ -53,10 +53,19 @@ const validationSchema = Yup.object().shape({
   hours: Yup.array().of(HourSchema),
   twitter: Yup.string()
     .matches(
-      /^https?:\/\/(www\.)?(twitter\.com|x\.com)\/.*/,
-      "Invalid URL, e.g. 'https://twitter.com/ or https://x.com/'"
+      /^@([a-zA-Z][a-zA-Z0-9_]{0,14}|[0-9][a-zA-Z0-9_]{0,14})$/,
```
hanapotski, line 56: not sure if we want to save the `@` in the db. I tried going to a twitter profile with it and it redirects to the one without it 🤔

client/src/components/Admin/OrganizationEdit/ContactDetails.js
```diff
@@ -65,36 +66,56 @@ export default function ContactDetails({
...
               onBlur={handleBlur}
               helperText={touched.twitter ? errors.twitter : ""}
               error={touched.twitter && Boolean(errors.twitter)}
+              InputProps={{
+                startAdornment: (
+                  <InputAdornment position="start">
+                    https://twitter.com/
+                  </InputAdornment>
+                ),
+              }}
```
hanapotski, line 91: instead of `startAdornment`, I was thinking of something like this. We might need to create a custom component for it 😅 
Screen Shot 2024-08-12 at 9 53 55 PM

monicakrystal: Screenshot 2024-08-13 at 10 00 12 PM
I did the twitter similar to your inspiration image and based on the conversation we had and kept the pinterest field the original suggestion from the stand up. Which one do you prefer?

monicakrystal: I can also change the pinterest field to have less empty space between the url and the handle.

client/src/components/Admin/OrganizationEdit.js
monicakrystal: yes, I believe both @username and username are accepted so I wasn't sure how people were going to write it. if you want without the @ symbol I can fix it to not accept it and only accept just the letters.

client/src/components/Admin/OrganizationEdit.js
```diff
@@ -53,10 +53,19 @@ const validationSchema = Yup.object().shape({
...
+      "Valid Twitter username required."
     )
-    .required("Full Twitter/X URL is required."),
+    .required("Valid Twitter username is required."),
```
hanapotski, line 59: I don't think these fields should be required 🤔 they should be optional 👌

client/src/components/Admin/OrganizationEdit/ContactDetails.js
hanapotski: nice! I think both could work and yeah maybe less space between the url and the handle 👌

client/src/components/Admin/OrganizationEdit/ContactDetails.js
hanapotski: since we're already adding the url, the tooltip should probably just say that it should be a valid username. Here's what's on pinterest [help center](https://help.pinterest.com/en/article/edit-your-profile#:~:text=For%20example%2C%20Pinterest%20has%20the,between%20three%20and%2030%20characters)

client/src/components/Admin/OrganizationEdit.js
monicakrystal: okay, i fixed the validation to make it optional and also changed the design for both the pinterest and twitter.
Screenshot 2024-08-21 at 4 24 54 PM

client/src/components/Admin/OrganizationEdit/ContactDetails.js
hanapotski: I think we can update the tooltip to something like `Must be a valid handle without the url`

client/src/components/Admin/ui/Label.js
```diff
@@ -57,7 +59,7 @@ const Label = ({ id, label, tooltipTitle, href }) => {
             sx={{ textTransform: "none" }}
             rel="noopener"
             component={Link}
-            href={href}
+            href={combinedLink} // Use the combined URL here
```
hanapotski, line 62: I think we need to update the check we're doing before we render this `Go To Link` button. Instead of just `href`, we should probably check for `handle`

client/src/components/Admin/OrganizationEdit/ContactDetails.js
```diff
@@ -1,6 +1,59 @@
...
+  },
+  "& .MuiInputAdornment-root": {
+    backgroundColor: "#F4F6F8",
+    padding: "22px 8px 22px 8px",
+    borderRight: "1px solid grey",
+    borderRadius: "4px 0 0 4px",
+    color: "#B0BEC5",
+    margin: "-20px 0",
+  },
+});
```
hanapotski, line 28: since we're not reusing this I would just inline the styles in a `<TextField />`
```const CustomTextFieldComponent = ({
  id,
  name,
  placeholder,
  value,
  onChange,
  onBlur,
  touched,
  errors,
  startAdornment,
}) => (
  <TextField
    id={id}
    sx={{
      "& .MuiOutlinedInput-root": {
        borderRadius: "4px",
        backgroundColor: "white",
        paddingLeft: "0",
        "& fieldset": {
          borderColor: "grey",
        },
        "& input::placeholder": {
          color: "#B0BEC5",
          opacity: 1,
        },
      },
      "& .MuiInputAdornment-root": {
        backgroundColor: "#F4F6F8",
        padding: "22px 8px 22px 8px",
        borderRight: "1px solid grey",
        borderRadius: "4px 0 0 4px",
        color: "#B0BEC5",
        margin: "-20px 0",
      },
    }}
    name={name}
    placeholder={placeholder}
    value={value}
    onChange={onChange}
    onBlur={onBlur}
    helperText={touched ? errors : ""}
    error={Boolean(touched && errors)}
    InputProps={{
      startAdornment: startAdornment && (
        <InputAdornment position="start">{startAdornment}</InputAdornment>
      ),
    }}
  />
);
```

monicakrystal: What do you mean by this? I thought by default it is using that style? If i use inline style like that I think i would have to include that inline style in every textfield ?

monicakrystal: okay, fixed both to check if theres a handle and the inline styles like you said above. Let me know if you need any changes to to.

client/src/components/Admin/ui/Label.js
```diff
@@ -50,14 +52,14 @@ const Label = ({ id, label, tooltipTitle, href }) => {
           )}
         </Stack>
 
-        {href && (
+        {href && handle && (
```
hanapotski, line 55: I think we can just check for `handle`, we'll always be passing the href right?

client/src/components/Admin/OrganizationEdit.js
```diff
@@ -52,11 +52,20 @@ const validationSchema = Yup.object().shape({
...
+    .nullable(),
+
+  pinterest: Yup.string()
+    .transform((value, originalValue) =>
+      originalValue.trim() === "" ? null : value
+    )
     .matches(
-      /^https?:\/\/(www\.)?(twitter\.com|x\.com)\/.*/,
-      "Invalid URL, e.g. 'https://twitter.com/ or https://x.com/'"
+      /^[a-zA-Z0-9.]{3,15}$/,
```
SAUMILDHANKAR, line 64: @hanapotski Thanks a lot for all your suggestions. 
@monicakrystal Great work on this PR, just wanted to share a few observations based on the [link](https://help.pinterest.com/en/article/edit-your-profile#:~:text=For%20example%2C%20Pinterest%20has%20the,between%20three%20and%2030%20characters) shared by Hannah:

1. The username allowed by the Pinterest policy can be 30 characters long and the current expression throws an error after 15.
2. Underscores are allowed by Pinterest whereas the dot is not.
3. Username can not be just made of the numbers however the above expression does not return an error if the username entered is just the numbers. I haven't figured out how to implement this yet, will share if I find anything. We can also discuss this during the meeting if this could be implemented later.

Thank you.

PR #2185, opened 2024-08-11
Added Twitter Validation and Tool tip #2184

### LINKED ISSUE #2184: Add client-side validation and tooltip for Twitter contact detail field
### Overview
We would like the data in our contact field to be accurate. In this issue, we will add client-side validation for the Twitter contact field.

### Action Items
- [] Add client-side validation which prompts user in case of an error. 
- [] Add Tooltip instructions
- [] Test it in your dev environment.

### Resources/Instructions
- Related file: https://github.com/hackforla/food-oasis/blob/develop/client/src/components/Admin/OrganizationEdit/ContactDetails.js

monicakrystal:
Closes #2184 

I added validation and a tool tip for the twitter field in the contact details.

Since twitter is now called X, I made it both accept http://twitter.com/ or http:// x.com/ and it will accept any username they enter after.

it also accepts http:// or https:// and www is optional. 

My question would be if it must start with https:// or http:// or can they write 
twitter.com/username or x.com/username ?

I would need to change it, if that is an option you guys want to add

Thank you!

here is a video that explains what I have written above. sound on to hear explanation or you can just view the video 

https://github.com/user-attachments/assets/a7e56208-32cf-4e31-8452-958d3c3476a5

### DIFF
--- a/client/src/components/Admin/OrganizationEdit.js
+++ b/client/src/components/Admin/OrganizationEdit.js
@@ -51,6 +51,12 @@ const validationSchema = Yup.object().shape({
   longitude: Yup.number().required("Longitude is required").min(-180).max(180),
   email: Yup.string().email("Invalid email address format"),
   hours: Yup.array().of(HourSchema),
+  twitter: Yup.string()
+    .matches(
+      /^https?:\/\/(www\.)?(twitter\.com|x\.com)\/.*/,
+      "Invalid URL, e.g. 'https://twitter.com/ or https://x.com/'"
+    )
+    .required("Full Twitter/X URL is required."),
   selectedCategoryIds: Yup.array().min(
     1,
     "You must select at least one category"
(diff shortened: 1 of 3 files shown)


client/src/components/Admin/OrganizationEdit.js
```diff
@@ -51,6 +51,12 @@ const validationSchema = Yup.object().shape({
   longitude: Yup.number().required("Longitude is required").min(-180).max(180),
   email: Yup.string().email("Invalid email address format"),
   hours: Yup.array().of(HourSchema),
+  twitter: Yup.string()
+    .matches(
+      /^https?:\/\/(www\.)?(twitter\.com|x\.com)\/.*/,
```
hanapotski, line 56: I think http gets redirected to https but we should probably just check against https to be consistent 🤔

hanapotski (approved):

PR #2204, opened 2024-09-11
2190 change css for organization name links to be more obvious

### LINKED ISSUE #2190: Change CSS for organization name links to be more obvious
### Overview

It's not obvious that each organization has a detail screen because links to it aren't obvious.

Many other food pantry websites have the same information that we have on the search results screen but nothing more.

Our site has much, much more than those sites, but people won't know this unless there are obvious links to bring them deeper in.

To fix this, we should make the organization names on the search results very obvious links.

To start, let's change the Org Name Links so that they underline all the time.  

Why all the time and not just on hover?  Because there's no hover effect on mobile screens.  Things need to look clickable right off the bat.

image

Note that on the org detail screen, the name of the org is not a link and will look like just a header, which it is.

If this is too jarring, we can try other methods via simple CSS changes and the dev and designer can try out things together -- the point is that there needs to be some obvious differentiation between what's a link and what's a label.

### Discussion

I think this should be a simple CSS update.

monicakrystal:
Closes #2190 

added underline to title, to make it look more clickable as requested
Screenshot 2024-09-10 at 5 57 06 PM
after

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/StakeholderPreview/StakeholderPreview.js
+++ b/client/src/components/FoodSeeker/SearchResults/StakeholderPreview/StakeholderPreview.js
@@ -200,8 +200,7 @@ const StakeholderPreview = ({ stakeholder, onSelect, isDesktop }) => {
               align="left"
               fontWeight="bold"
               sx={{
-                color: { xs: "black", md: "inherit" },
-                textDecoration: { xs: "underline", md: "none" },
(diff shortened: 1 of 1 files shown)


ExperimentsInHonesty:
@monicakrystal Please resolve conflicts on this pr

junjun107 (approved):
LGTM

PR #2223, opened 2024-09-23
Adding settimeout to enable slower zip code typing

### LINKED ISSUE #2186: Bug: Sometimes the map shows incorrect area after typing a zipcode into search bar on landing page.
- [ ] Please assign @fancyham  and @Shienny1 to the PR

#### Describe the bug 

Sometimes the map shows incorrect area after typing a zipcode into search bar on landing page.  

This happens only sometimes!

In this video (turn on audio), taken at the beginning of July 2024:

https://github.com/user-attachments/assets/956951e0-00f2-4583-a6a4-60329e6c8489

1. User typed in "90013"
2. Map shows an address in 90045 zip code (as seen in search address field at top of screen)

SAUMILDHANKAR:
- Closes #2186 
- Adding settimeout just makes sure that the user is not able to type the zip code too fast (and so the zipcode resolves correctly). Please let me know if this might not be an acceptable solution and I can look for some other way to fix this.

### DIFF
--- a/client/src/components/FoodSeeker/AddressDropDown.js
+++ b/client/src/components/FoodSeeker/AddressDropDown.js
@@ -31,7 +31,9 @@ export default function AddressDropDown({ autoFocus }) {
   const handleInputChange = (delta) => {
     if (!delta) return;
     const safeValue = typeof delta === "string" ? delta : delta.target.value;
-    setInputVal(safeValue);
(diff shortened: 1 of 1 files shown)


SAUMILDHANKAR:
The solution in this PR makes typing in the address field difficult. So, I will work on reverting the changes of this PR next.

Issue #2219, opened 2024-09-17
Serious Bug: Map - Mobile map pans incorrectly when viewing an organization


fancyham:
VERY HIGH PRIORITY -- HOT FIX REQUESTED.

### Progress

- [x] Create partial-solution to get it mostly working again #2229
- [x] Push hot fix to public
- [x] Create solution where map pans to new location **without** resetting zoom level (please assign @fancyham and @Shienny1 to PR to review)
- [x] Push to public

(~Sept 15, 2024)
Re-creating this as a new issue per @hanapotski because the solution for 
- #2028 
didn't seem to work -- the issue with the map panning to the wrong location still exists.

A hint for this: Part of the issue seems to be the map's zoom level.  At default zoom, it's not a problem.  Changing the zoom causes the problem.

#### Describe the bug 

Serious bug:

When viewing a list of organizations, and tapping one to get details about it, the map does not update show or highlight the organization’s pin.  Often the pin is way off-screen.

The map should re-center (ideally with a smooth scroll/pan) on the highlighted pin automatically, ideally showing both the user’s current location and the pin so that they can determine the relative location.

#### Steps to reproduce the issue

(if applicable, please specify platform (iOS, Android, Windows, Mac version) and brower)

1. Go to foodoasis.net
2. Enter a location to get a list of organizations
3. Scroll around the map, **zoom in or zoom out**
4. Tap an organization to view the organization detail page
5. Map does not update to show the organization’s pin.

See more videos at
-#2028

#### What's the expected result?

Map should re-center on the organization’s pin automatically, each time a new organization is viewed.  

Ideally, the map smoothly pans to the new location, and ideally, the user’s current location / entered address is also visible on the map (though this may be difficult if the user has panned the map manually)

See Apple or Google maps or Yelp on mobile phones for example.

#### What's the actual result?

The organization’s pin may be highlighted, but is often off-screen with no indication for how to find the pin itself.

This is a regression — the previous version centered on highlighted pins properly.

#### Additional details / screenshot

A similar problem appears when searching from the map screen. 
1. From landing page, enter an address. On desktop, map Pin appears in the wrong place but map pans correct place at first.
![IMG_1253](https://github.com/user-attachments/assets/cf8cc736-3d32-4d47-b1df-c11b90211b6e)

2. Do a search again from the map page and the map now re-centers in the wrong place, and the pin doesn’t move to the correct place. 
[...]

PR #2229, opened 2024-09-27
Reset zoom when flying to a new location

hanapotski:
Addresses #2219 

https://github.com/user-attachments/assets/ac7f3e50-4476-42a3-a106-c11215839276

### DIFF
--- a/client/src/hooks/useMapbox.js
+++ b/client/src/hooks/useMapbox.js
@@ -1,7 +1,8 @@
+import { defaultViewport } from "helpers/Configuration";
 import { useRef } from "react";
+import { useMap } from "react-map-gl";
 import { useListPanel } from "../appReducer";
 import useBreakpoints from "./useBreakpoints";
-import { useMap } from "react-map-gl";
 
 export const useMapbox = () => {
   const mapRef = useRef();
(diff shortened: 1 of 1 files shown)


junjun107 (approved):
Great idea to reset the zoom. 
LGTM

hanapotski:
> Great idea to reset the zoom. LGTM

that was your idea 🤭 I was thinking of something else but this is easier lol 🙌

fancyham:
**Hi - this is a clever fix for now and if we can push this as a hot-fix, that would be terrific*

Long-term, though, we should still figure it out a solution such that the zoom stays the same so that weird/frustrating things don't happen (for example, if you look at 0:16 on the video, clicking on a cluster of locations starts an unintended zoom, which prevents the user from clicking on other items in the cluster, because they've disappeared at the default zoom level).  We should keep the map's zoom state what the user last chose when zipping between locations for this reason.

Also, I'm not sure how this works for the original issue's open/closed status -- it doesn't actually fix the problem mentioned there but it but it does make a huge step  (i.e. pan to location without the zoom), especially if we can push the hot fix.

Also, I noticed in the video that the first autocomplete entry is 90045... I think that's a hint at why we keep jumping to 90045 if the user types in an address and hits return quickly (before autocomplete finishes)
- https://github.com/hackforla/food-oasis/issues/2186

Issue #1692, opened 2023-05-18
Bug: User-submitted text address doesn’t properly resolve into an address or GPS,  causing a failed search


fancyham:
### Overview

Bug discovered while doing 
- #1668 
…Entering an address or zip code into the search text field (then hitting ‘return’, not using autocomplete) often fails to resolve to a location on the map, resulting in ‘zero results’, but then a dead-end.  We should provide similar addresses when needed, and at least resolve zip codes properly.

This is also related to
#1115
where we would want people to be able to search for org names and addresses using the search box.

### To replicate

1. Go to http://foodoasis.la
2. Type in 90045 (or nearly any other address), press enter (do not use the autocomplete)
3. The problem: The zip code is not resolved to a location on the map, and there are zero results.

It seems to be that right now, the only way to get an address on the map is to choose one from autocomplete or to use the “locate me” GPS function.

### Desired result

- [ ] After typing in an address or zip code that doesn’t resolve to an address or GPS location, show an error or a list of possible addresses.  See Google Maps for an example of incomplete-address resolution:
- [ ] (if possible) also support searching for known business and organization names (see #1115)

<img width="528" alt="image" src="https://github.com/hackforla/food-oasis/assets/3376957/f6c85d2f-3c22-4980-90de-5d5abc283af4">

If this type of address resolution is not possible, we need to explicitly tell people that they must choose an autocomplete option, but this breaks with expected behavior and we would want to fix this later.

fancyham:
This seems to have been resolved, so closing.

fancyham:
Reopened as this contains a good flow for resolving malformed or incorrect addresses, and we do need a solution for this particular error case, as we discover more edge cases.

- #2186

fancyham:
Getting there — it doesn't go to the wrong zipcode — but try this link — https://devla.foodoasis.net/organizations?address=90015 (which is what I got when I typed really fast on the landing page.  After that, perhaps the results were cached and came up too fast).

fancyham:
Also see
- #2129

VirginiaWu11:
Hey @fancyham, I believe this was solved in develop by preventing the users from hitting search before the response was received from the server.

PR #2235, opened 2024-10-07
changed geocoder autocomplete to false

### LINKED ISSUE #2186: Bug: Sometimes the map shows incorrect area after typing a zipcode into search bar on landing page.

SAUMILDHANKAR:
- Fixes #2186 
- autocomplete prompts only appear if there is an exact match.
- hitting the return key really fast still breaks the search, not getting 90045 instead of 90015 though.

https://github.com/user-attachments/assets/ec2b2c20-b73a-437f-b63b-567cd3222d0b

### DIFF
--- a/client/src/hooks/useMapboxGeocoder.js
+++ b/client/src/hooks/useMapboxGeocoder.js
@@ -57,7 +57,7 @@ export function useMapboxGeocoder() {
           : tenantId === 3
           ? hawaiiLatLong
           : losAngelesCountyLatLong;
-      const mapboxUrl = `${baseUrl}/${searchString}.json?bbox=${bbox}&access_token=${process.env.REACT_APP_MAPBOX_ACCESS_TOKEN}`;
+      const mapboxUrl = `${baseUrl}/${searchString}.json?bbox=${bbox}&autocomplete=false&access_token=${process.env.REACT_APP_MAPBOX_ACCESS_TOKEN}`;
 
       dispatch({ type: actionTypes.FETCH_REQUEST });
       try {


fancyham:
Hi, this (the video above) is definitely an improvement!… but now feels broken in a different way :)

The user quickly typing 90015 and hitting return… that should be a valid input, so us going to the map and showing no address in the field (as in the video) feels broken but in a different way.  Like the user entered something and we ignored it. 

(I can see in the URL that the 90015 is in the request, too… that should be a valid input, no?)

So, this is a tough piece as you can see from past issues around this: https://github.com/hackforla/food-oasis/issues?q=is%3Aissue+autocomplete+is%3Aclosed

The desired flow is in this issue (it was iceboxed and then I closed it but now realize that I should have left it open) 

https://github.com/hackforla/food-oasis/issues/1692

Describes how we should try to resolve what they typed in - perhaps autocomplete happens silently before displaying the map, if needed… and if a match isn’t found, then we display an error or ask user to disambiguate. 

So…
1. user enters 90015, hits return too quickly for autocomplete to respond 
2. System should silently look for autocompletes for 90015
3. If it finds autocompletes, resolve to the top hit and use that as the address.  END. 
4. If it does NOT find an autocomplete that makes sense, we should display an error message  “could not find address” error text, next to the address field which contains the user’s input.  User can then re-enter an address.
  
Might that be possible?  

Let me know if a sketch would be helpful. 

Btw, the video you made was very helpful!

SAUMILDHANKAR:
Hi Bryan, thanks a lot for your review and sharing the desired flow. I had to take a few sick days off last week so didn't make much progress but planning to work on it this week and share update asap. Thanks a lot for your help.

SAUMILDHANKAR:
Handed off the issue to Hannah, so closing this PR.

PR #2236, opened 2024-10-07
maintains zoom level and centers active pins #2219

### LINKED ISSUE #2219: Serious Bug: Map - Mobile map pans incorrectly when viewing an organization

whitneywind:
Fixes #2219 

- These changes update the flyTo function so that it keeps the zoom level consistent with the user's current zoom level instead of reverting back to the original zoom level
- They also calculate the longitude and latitude offsets as needed to keep the pins centered in the visible part of the map 

https://github.com/user-attachments/assets/21aa01de-4c85-4842-b2bf-a2f6dc765c7c

https://github.com/user-attachments/assets/147c8f8d-a7f2-436b-a289-e8328026fc10

### DIFF
--- a/client/src/hooks/useMapbox.js
+++ b/client/src/hooks/useMapbox.js
@@ -1,4 +1,3 @@
-import { defaultViewport } from "helpers/Configuration";
 import { useRef } from "react";
 import { useMap } from "react-map-gl";
 import { useListPanel } from "../appReducer";
@@ -28,12 +27,28 @@ export const useMapbox = () => {
     if (!mapbox.default) {
       return;
     }
+
+    // gets the user's current zoom level
+    const currentZoom = mapbox.default.getZoom();
+
+    // calculates longitude offset according to zoom level for panel open desktop
+    const baseLongOffset = 0.08;
+    const longitudeOffset = baseLongOffset * Math.pow(2, 11 - currentZoom);
+
+    // calculates latitude offset for mobile according to zoom level and takes screen height into account
+    const baseLatOffset = 0.05;
+    const screenHeight = window.innerHeight;
+    const referenceHeight = 800;
+    const screenHeightFactor = screenHeight / referenceHeight;
+    const latitudeOffset =
+      baseLatOffset * Math.pow(2, 11 - currentZoom) * screenHeightFactor;
+
     mapbox.default.flyTo({
       center: [
-        isListPanelOpen && isDesktop ? longitude - 0.08 : longitude,
-        isMobile ? latitude - 0.04 : latitude,
+        isListPanelOpen && isDesktop ? longitude - longitudeOffset : longitude,
+        isMobile ? latitude - latitudeOffset : latitude,
       ],
-      zoom: defaultViewport.zoom,
+      zoom: currentZoom,
       duration: 2000,
     });
   };


fancyham:
Looks nice!  

I can’t see the code changes that you made, but hopefully nothing is hardcoded and the locations and pans and everything are being calculated dynamically based on the visible map area? ( which is tricky because the tax panel can be different sizes and the user can adjust the panel size manually)

Please also test that it works correctly with all map/text panel splits on mobile and with all screen sizes!

The one situation that I’m thinking of here is where the user is on mobile and has adjusted the split screen so that the map is ke 80% of the screen and the text panel is 20%. Does the map centering also work there?  

If it does, then this looks good to me and I’d say push as fast as possible to public if @Shienny1 doesn’t find any problems.

whitneywind:
@fancyham That's a good point! Right now it doesn't account for a user manually adjusting the panel sizing after they make a search. I'll update it to factor in the panel size when calculating the long. and lat. offsets.

whitneywind:
@fancyham Okay, so it does stay horizontally centered no matter what. The vertical positioning doesn't change on mobile when the text panel is pulled down. Is that the desired behavior? Or should the pin and visible map area also move vertically when the text panel is pulled up and down?

fancyham:
It'd be amazing if adjusting the map's percentage of the screen kept the center of the map centered dynamically as the user adjusts the split.  The zoom level stays the same.  Would that be possible?  Sounds hard but I think that's what a user would ideally want to happen.  

This would help users keep track of what they're looking at, rather than be confused by the pin they're looking at getting covered up by the text panel they just increased the size of, then having to manually pan the map to find the pin again.

whitneywind:
@fancyham Please let me know if I'm understanding correctly! So the vertical centering should change with the panel resizing, unlike it is now? If so, the only problem I see is that the map area would need to change with each adjustment in order to keep the pin centered. And since the user is probably manually moving around the map when it's 80% of the screen, if they moved to a different area and then adjust the panel, the map scrolling up and changing to account for that might be confusing. 

Right now it's the same as it is on my mobile yelp app or google maps app, where if they move the panel up or down the map position stays the same to help the user stay oriented.

entrotech (approved):
As far as I can tell, this is working really well! Tried various zoom levels, and various ways of causing the panning (typing in the address box, clicking on map pins, selecting a listing from the list box), and it all seems to work really well.  I think it might be kind of jarring for the user to pan the map as the user drags the drag bar up and down on mobile or opens/closes the side drawer on desktop.

fancyham:
Good points!  I think the map dynamically keeping things centered is perhaps too clever for its own good, and yes, it could be visually jarring.  Knowing that Apple and Google don’t do it are great datapoints.

This is good to push!  Thanks!

fancyham:
BTW, I found an example of the ‘automatically zoom to keep what you were looking at on-screen even as you change the viewport’ idea we were discussing above inside Uber’s app.  

It looks tricky to implement but quite useful — in this case, I’m looking at going to an airport — so it keeps that full route on-screen by shrinking or expanding it as needed, meaning that I’m comforted in knowing that Uber hasn’t forgotten the route I’ve been working on (imagine what it’d fee like if the map didn’t zoom… people trust what they can see, so if something important goes off-screen, that can be a problem):

https://github.com/user-attachments/assets/c4bf85ac-28d7-4e56-92a8-ea447146cde8

PR #2311, opened 2024-12-17
button component for correct zoom #2299

### LINKED ISSUE #2299: Bug: Map’s zoom buttons causes undesired panning
#### Describe the bug 

Clicking on the on-map +/- zoom buttons cause the map to zoom in and out, however the zoom’s origin/center is in the wrong place, causing the map to pan to the side when zooming in and out.

1. On Desktop, go to the map (this might also be a problem on mobile as well) by typing in “90015”. The map will appear with a red pin in the center.
2. Click on the “+”

The red pin should stay centered on the map while the zoom level increases.

The red pin slides to the side, even though it’s in the center of the visible map.

The center of the map seems to be calculated by including the hidden portion under the text panel.

The center should be calculated based on the currently visible/exposed map area instead.

#### Device configuration
- Device: Mac with latest 
- OS version: The latest as of today
- Browser: Safari

whitneywind:
Closes #2299 

This is a custom button component replacing the default zoom buttons in Mapbox. Mapbox's default button behavior isn't customizable, so the buttons in this replacement component account for the fact that our map's marker's are on the right side in desktop view when the panel is open, and adjust accordingly. The mouse and trackpad scrolling is unaffected.

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.jsx
@@ -7,7 +7,7 @@ import { useCallback, useEffect, useState } from "react";
 // recommendation from Mapbox team
 // https://github.com/mapbox/mapbox-gl-js/issues/10173  See comment by IvanDreamer on Mar 22
 // for craco.config.js contents
-import { Grid } from "@mui/material";
+import { Grid, Box } from "@mui/material";
 import { MAPBOX_STYLE } from "constants/map";
 import { MAPBOX_ACCESS_TOKEN, DEFAULT_VIEWPORT } from "helpers/Constants";
 import useBreakpoints from "hooks/useBreakpoints";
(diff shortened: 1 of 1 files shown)


SAUMILDHANKAR (approved):
Zoom looks working fine to me. Thanks Whitney.

client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.jsx
```diff
@@ -134,6 +137,92 @@ const ResultsMap = ({ stakeholders, categoryIds, toggleCategory, loading }) => {
...
+          sx={buttonStyles}
+        >
+          ＋
+        </Box>
+        <Box
+          onClick={handleZoomOut}
+          sx={[buttonStyles, { fontSize: 14 }]}
+        >
+          ―
+        </Box>
```
hanapotski, line 221: instead of <Box/>, I think we can use <IconButton /> and just style it

hanapotski (approved):

PR #2312, opened 2024-12-31
Fixes issue where scrollbar was appearing on safari mobile

### LINKED ISSUE #2300: Bug: On my Mac, scroll bars appear under the Filters buttons
#### Describe the bug 

On my Mac, I have my scroll bars set to “always appear” — meaning that scroll bars are visible by default (like in the old days) rather than appearing during scrolling only.

However, I think that when I test food oasis in a tiny window to simulate a mobile device, these scroll bars are appearing under the scrollable filter buttons (which extended beyond the edge of the screen)

I suggest using a CSS setting to hide scroll bars for the filter buttons.

This may not be a common case, but it ends up looking weird.  

1. In Safari on Mac, visit la.foodoasis.net, then go into Developer Responsive Screen Size simulator (Shift-command-R, I think)
2. Shrink the viewport to a small mobile phone size

The map should look normal, like it does on an iPhone.

What actually happens is that there’s a scroll bar shadow appearing under the Filter buttons.

- Screenshot

Screenshot 2024-11-25 at 5 53 44 PM

#### Device configuration
- Device: Mac
- OS version: Latest
- Browser: Safari

danvgar:
Fixes #2300 

Added additional browser compatibility for removing scrollbar.

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsMap/ResultsMap.jsx
@@ -185,10 +185,14 @@ const ResultsMap = ({ stakeholders, categoryIds, toggleCategory, loading }) => {
           display="inline-flex"
           alignItems="flex-start"
           sx={{
-            overflow: "auto",
+            overflowX: "auto", 
+            overflowY: "hidden", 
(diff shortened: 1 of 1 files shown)


danvgar:
Oh woops, I forgot to rebase this. Let me know if you'd like me to

PR #2328, opened 2025-01-21
Add more info link

### LINKED ISSUE #2304: Add a "more info…" link to org details for each search result (because people don't realize there's more info)
### Context

In the recent redesign, we simplified the search results list, and removed the “more info” button. 

We saw with some small tests, that people didn’t realize there even was “more info” for each listing. 

To make this more obvious, we made the org names underlined to look like links #2190 

But that is probably not enough. 

Let’s make it more obvious that there’s often _lots_ more info for each listing. 

### How is it now?

There's no obvious indicator that there's a lot more information available for each listing on the search results.

image

In 2020, we used to have a separate "more info" button, next to the "call" and now-gone "share" button.

 
### Proposed new feature
- Add a “more info…” underlined/colored text hyperlink link at the bottom of each listing’s description, above the directions and call buttons, that links to that listing's detail screen (same as if you clicked the name of the organization).  Ideally, this is a deep-link URL (so user can right-click and share it).

(please use the same font and color as More Hours")

### Discussion
What are some other possible solutions?
- Considered adding a “More info” button to each listing (after the directions, call buttons) like we used to have back in 2020, but I suspect that people didn't realize it was there, and it ended up taking additional vertical space. Let's build the text link first.  Adding a "Share" button may be better than a "More info" button

hanapotski:
Closes #2304

Screen Shot 2025-01-20 at 8 21 00 PM

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/StakeholderPreview/StakeholderPreview.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/StakeholderPreview/StakeholderPreview.jsx
@@ -3,8 +3,9 @@ import {
   Button,
   Chip,
   Divider,
+  Link,
   Stack,
-  Typography
+  Typography,
 } from "@mui/material";
 import Grid2 from "@mui/material/Unstable_Grid2";
 import {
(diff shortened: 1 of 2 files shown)


SAUMILDHANKAR (approved):
Thanks @hanapotski, both the links looking good to me.

PR #2354, opened 2025-02-13
Fix spacing

### LINKED ISSUE #2352: Search results and details: reduce extraneous padding and spacing of the icon and text for more compactness
### Context
We want to remove some of the extraneous padding on the text views on search results and listing detail screens to increase information density.

Please check on Chrome and Safari — the rule for the detail listing screen didn't seem to be the same length when we were playing with it during our meeting?

keywords: margin, padding, left edge

### Proposed change

Showing the current as-built but moving the left edge of the search results over.   It'll end up looking similar/same as the margins on the mobile view (the desktop's search results are supposed to be like a taller mobile view)

The goal is: "ah, that feels right and comfortable" and not "what's with this extra space on the left?"

It ideally aligns with the left edge of the logo.

- [ ] The "Locate free food in …" subhead should also be left-aligned

image

See below for notes on 

- [ ] the scroll bar's thumb (should be on right edge — it is on my machine, but in the version you and I were tweaking at the last meeting, the thumb was further to the left with a margin to the right)
- [ ] Change the row hover from a drop-shadow to a color

The current metrics of the as-built are different, but here's what we're trying to align (this is from the 2024 redesigns)
image

OLD sketch:
Image

- Remove the extraneous padding on the left.
- Eyeball it so that it matches the padding/margin on the right.  The icon's shape may make it need to be a slightly different numerical margin to make it 'look and feel' like the left and right margins are the same.  (it's an old graphic design / typography trick that exact measurement matter less than how it feels visually)

- [ ] Confirm that the "< Back to results" and horizontal rule match the new margins on all browsers

- [ ] On mobile: Move the horizontal rule under "< Back to results" so it matches the desktop... it's there to show where the scrolling region ends.
- [ ] On mobile: Add back the # of results, and add some new text "(scroll down)" to help people realize that there's a ton of listings if they scroll... it's not particularly obvious now.

- [ ] Please shoot screenshots when you're feeling comfortable.
- [ ] There might be different sets of padding/margin that feel right — some might be more compact than others.  Feel free to send screenshot to design / share variations to review.  

### Discussion

The goal is to let users: feel comfortable, not crowded, at-ease, not pressured.  But also not so spacious or information-sparse that we're wasting people's time.

So effective and informative without feeling burdensome or oppressive or scary.

### References:
For reference, here is how it used to look:

Image

hanapotski:
Closes #2352 

https://github.com/user-attachments/assets/fa5559ed-429c-42d1-8ee3-3420d17e9e6b

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsList/ResultsList.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsList/ResultsList.jsx
@@ -31,106 +31,90 @@ const ResultsList = ({ stakeholders, loading, handleReset }) => {
     analytics.postEvent("showList");
   }, []);
 
-  const renderView = () => {
-    if (selectedOrganization)
-      return (
-        <StakeholderDetails
-          onBackClick={() => dispatch({ type: "RESET_SELECTED_ORGANIZATION" })}
-          isDesktop={isDesktop}
-        />
-      );
-
-    if (loading) {
-      return (
-        <Stack justifyContent="center" alignContent="center">
-          <CircularProgress />
-        </Stack>
-      );
-    }
-
+  if (selectedOrganization)
     return (
-      <>
-        {stakeholders.length > 0 && isDesktop && (
-          <Stack sx={{ width: 1, padding: "0 35px 0 65px" }}>
-            <Typography
-              sx={(theme) => ({
-                paddingTop: "1.4rem",
-                textAlign: "left",
-                fontWeight: "bold",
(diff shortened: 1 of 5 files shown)


hanapotski:
@fancyham what do you think?

fancyham:
Pretty good but a few adjustments! — I added some mockups showing alignment to the parent issue here:
- #2352

SAUMILDHANKAR (approved):
Great improvements. Thanks Hannah.

PR #2360, opened 2025-02-17
Geolocate Button on Map #1725

### LINKED ISSUE #1725: Map: Move GPS button from top rows onto map for better usability
### Overview
Currently, the GPS button is next to the address field which is a pretty non-standard place nowadays, and we suspect users don't know it’s there.  Let's move the button onto the map itself, similar to Apple Maps and Google Maps.  

Something like this:
image

### Action Items
- [x] Devs investigate if we can move button onto map, let designers know
- [ ] Designers specify both location and icon for GPS button, and what to do with the address bar row once the existing button goes away.
- [x] Devs implement

### Resources/Instructions
- I noticed that MapBox's sample maps has this — hopefully it's easy for us to do.

whitneywind:
Resolves #1725 

Moves the geolocate button below the zoom buttons in the desktop version. In the mobile it stay on top to avoid overcrowding the map.


client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.jsx
```diff
@@ -29,14 +25,10 @@ const ResultsFilters = ({
...
-  const userCoordinates = useUserCoordinates();
-  const startIconCoordinates = searchCoordinates || userCoordinates;
   const locationPermission = useLocationPermission();
   const [error, setError] = useState("");
   const hasAdvancedFilterFeatureFlag = useFeatureFlag("advancedFilter");
-  const { flyTo } = useMapbox();
+  const { isMobile } = useBreakpoints();
 
   useEffect(() => {
     if (error && locationPermission === "granted") {
```
hanapotski, line 34: since we moved the error in the new component, I think it makes sense to move this there too 🤔

whitneywind: @hanapotski Good catch, thank you! Does it look okay with the commit I just made?

PR #2361, opened 2025-02-18
2351 add distance under icon

### LINKED ISSUE #2351: Feature: Add distance from address under the icons in list view
### Context
We want to show distance under our icons in list view again (it was removed during the 2024 redesign in order to simplify the look and feel, but it seems that it's information that people are looking for)

How is it used
- to understand the order and logic of the items in a search results
- to get a rough sense of distance to help filter out good and bad candidates.   2 miles is too far to walk, for example.
- it's important, but probably not as other information for that listing such as hours, etc.

### Proposed new feature

Image

- Try using the same text style / format / color as hours
- Try eyeballing the space between icon and text… it should feel natural.  Not too close, not too far.

- [ ] If you want a stretch goal, allow us to also display distance in km (in case Food Oasis ever goes international) when desired... that might need some global locale setting, which may not exist.  If it doesn't exist, ask @entrotech or tell me @fancyham and we will create an issue to track it for the future.

- [ ] Before finalizing:
  - [ ] Please send screenshot for designers to check out
  - [ ] Also send screenshot with variants of color — perhaps 66% black, 50% black.  I don't know if we have other standard colors, but try those out first, before creating a custom color.
  - [ ] Basically, we want to pick something that doesn't get in the way but conveys the info with enough visual contrast that it meets accessibility guidelines.

### References:
Here's what it used to look like in 2022:

Image

VirginiaWu11:
-  closes #2351

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.jsx
@@ -333,9 +333,14 @@ const StakeholderDetails = ({ onBackClick, isDesktop }) => {
               direction="column"
               justifyContent="flex-start"
               alignItems="center"
+              gap={2}
               sx={{ marginTop: ".2rem", marginRight: "1rem" }}
             >
               <StakeholderIcon stakeholder={selectedOrganization} />
+              <Typography variant="body2" align="center">
+                {Boolean(selectedOrganization?.distance) &&
+                  `${selectedOrganization.distance?.toFixed(1)} mi`}
+              </Typography>
             </Stack>
           </Grid2>
           <Grid2 xs={10}>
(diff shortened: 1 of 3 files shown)


client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.jsx
```diff
@@ -333,9 +333,14 @@ const StakeholderDetails = ({ onBackClick, isDesktop }) => {
               direction="column"
               justifyContent="flex-start"
               alignItems="center"
+              gap={2}
               sx={{ marginTop: ".2rem", marginRight: "1rem" }}
             >
               <StakeholderIcon stakeholder={selectedOrganization} />
+              <Typography variant="body2" align="center">
+                {Boolean(selectedOrganization?.distance) &&
```
hanapotski, line 341: wonder if we want to show this `distance` is `0` 🤔

hanapotski (approved):

PR #2379, opened 2025-03-03
2353 add sticky filter title on panel

### LINKED ISSUE #2353: Filters editing panel should have a sticky row at the top (similar to the detail screens).  Mobile + Desktop
### Context
On mobile, the filters panel's title and dismiss button can be scrolled off-screen, causing confusion
On desktop, the filter form elements should be scrollable but aren't

Let's lock those on-screen while the filter form itself can be scrolled, while the header and dismiss button floats

### How is it now?
Video of now (has audio):

https://github.com/user-attachments/assets/2737142b-fb10-44a5-84c8-04a67f608721
 
### Proposed new feature
- [ ] Lock the top row of the filters panel so that the title, dismiss button, and horizontal rule are sticky

Also!
On desktop:
- [ ] Title, dismiss button, and horizontol rule are sticky, and make sure the filters form can be scrolled when window is very wide and short (it currently is not scrollable, causing some elements to be unreachable at certain window sizes)

VirginiaWu11:
- Closes #2353

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.jsx
@@ -29,7 +29,7 @@ import {
   useNoKnownEligibilityRequirementsFilter,
 } from "../../../../appReducer";
 import { getDayTimeNow } from "helpers";
-import { useEffect, useState } from "react";
+import useHeaderHeight from "hooks/useHeaderHeight";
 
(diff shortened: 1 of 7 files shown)


VirginiaWu11:
- see if i can fix the gap on desktop.
- padding at the top make it less so its centered
- scrollbar should be under the sticky part

VirginiaWu11:
- fix widget edge case, overlapping footer 
- footer on refresh height is too tall

VirginiaWu11:
image

hanapotski (approved):

Issue #2344, opened 2025-02-02
Connect the endpoints to create and update the announcement table


SAUMILDHANKAR:
### Overview
This issue is related to issue #2326. In this issue, we will connect the front end to the database.

### Action Items
- [x] Connect the server routes to the database.
- [x] Run a test to make sure it works.
- [x] Once the PR associated with this issue is merged or this issue is closed, check issue #2326, if all its sub-issues are marked complete, move that issue to the prioritized backlog column so someone can pick it up and check if that issue is complete or needs more work.

### Resources/Instructions
- Helpful similar PRs https://github.com/hackforla/food-oasis/pull/2033, https://github.com/hackforla/food-oasis/pull/2043

ExperimentsInHonesty:
waiting for response to https://github.com/hackforla/food-oasis/issues/2343#issuecomment-2856760292

PR #2387, opened 2025-03-18
need to debug put request

### LINKED ISSUE #2339: Create a new table for announcements
### Overview
We need to add a table in our database for announcements.
We would be using update CRUD operation.

### Action Items
- [ ] Need to keep track of all old announcements or delete the old ones

REPLACE THIS TEXT -If the issue has already been researched, and the course of action is clear, this will describe the steps.  However, if the steps can be divided into tasks for more than one person, we recommend dividing it up into separate issues, or assigning it as a pair programming task.

### Resources/Instructions
REPLACE THIS TEXT -If there is a website which has documentation that helps with this issue provide the link(s) here.

SAUMILDHANKAR:
- Related to #2344

### DIFF
--- a/client/src/components/Admin/LocalNotifications.jsx
+++ b/client/src/components/Admin/LocalNotifications.jsx
@@ -1,0 +1,373 @@
+import {
+  Box,
+  Button,
+  Container,
+  Modal,
+  Paper,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TablePagination,
+  TableRow,
+  TextField,
+  Typography,
+} from "@mui/material";
+import { Formik } from "formik";
+import { TENANT_ID } from "helpers/Constants";
+//import { useTags } from "hooks/useTags";
+import { useLocalNotifications } from "../../hooks/useLocalNotifications";
+import { useEffect, useState } from "react";
+import { Navigate, useLocation } from "react-router-dom";
+import * as localNotificationsService from "../../services/localNotifications-service";
+import { IconButton } from "../UI/StandardButton";
+import Label from "./ui/Label";
+
+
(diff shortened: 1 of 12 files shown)


client/src/components/Admin/LocalNotifications.jsx
```diff
@@ -0,0 +1,364 @@
...
+import * as localNotificationsService from "../../services/localNotifications-service";
+import { IconButton } from "../UI/StandardButton";
+import Label from "./ui/Label";
+
+const columns = [
+  { id: 1, name: "march3" },
+];
+
+//const apiiii = await localNotificationsService.getAlllocalNotifications();
+const lanat = await localNotificationsService.update(4, {id: 4, name: "march" });
```
SAUMILDHANKAR, line 32: This is where I have been working at to update the table.
The server error I get is:
```
error: null value in column "name" of relation "local_notifications" violates not-null constraint
```

VirginiaWu11: Hey @SAUMILDHANKAR ,
@junjun107 and I looked into this and found that `null` error was shown because 
image
in the `localNotifications-service`, the `update` function is expecting 2 parameters, but the `localNotificationsService.update` function called in the `localNotifications-controller` is only passing in the id without the 2nd parameter of `name` causing name to be null
image

note: the id is autogenerated when a notification is created in the database, so when passing the id to update, the id should match the one in the database

SAUMILDHANKAR:
@VirginiaWu11 @hanapotski Thanks a lot for pair programming last week and helping me fix the PUT request. It is working now and I am planning to create a formal PR next with tests for each requests. Also, planning to close this draft PR. Hope that is ok. Thanks a lot for your help.

Issue #2382, opened 2025-03-04
Pop-Up Survey design update


VirginiaWu11:
### Overview
Update the design of Pop-Up Survey to match the figma design below.

### Action Items
Update the design of Pop-Up Survey 

![Image](https://github.com/user-attachments/assets/2823fd67-271d-4428-861e-6cd6fd6f22c2)

Figma: https://www.figma.com/design/9hPmNSsyaH2VMGNz5mffmI/FOLA-Design-%232?node-id=8417-41698&t=xztMUBTZbvJWRUkc-4

This dev issue was created from design issue #1769

VirginiaWu11:
Hi @ihorng,

On the dev meeting, some things came up regarding this [design](https://www.figma.com/design/9hPmNSsyaH2VMGNz5mffmI/FOLA-Design-%232?node-id=8417-41698&t=xztMUBTZbvJWRUkc-4) for the pop-up survey.

1. The default scaling factor for spacing on Material UI is 8px. A spacing for 1 would be 8px, spacing of 2 would be 16px (2*8px), etc. We have been using this default throughout the site, so using spacing which is divisible by 8 would make our code more consistent. Let me know your thoughts and if you'd like to update the figma to match this system.
![Image](https://github.com/user-attachments/assets/d3c95b11-f11a-4300-a301-1502211a1a8d)

2. We also noticed that the colors are from another palette on figma. Would it be okay to use the colors we currently have? Here is our current [colors](https://github.com/hackforla/food-oasis/blob/develop/client/src/theme/palette.tsx) 

screenshot for reference:

<img width="221" alt="Image" src="https://github.com/user-attachments/assets/703f2a9a-b95e-446f-a68b-39f8fc537210" />


cc @hanapotski @fancyham

ihorng:
@VirginiaWu11 Thank you for sending me the info, I will take a look and update the pop up survey designs.

Would you happen to have a screenshot of all the call to action buttons that are being used right now on the website?

VirginiaWu11:
Hey @ihorng, we have https://devla.foodoasis.net/admin/muidemo this link that shows some of our components.

ihorng:
@VirginiaWu11 

1. See attachment. Are the links shown visual representations of the states described in the document?

<img width="1039" alt="Image" src="https://github.com/user-attachments/assets/651b8fb3-2d3c-4a22-81f7-9a315b970ead" />


&gt;

[...]

PR #2390, opened 2025-03-24
update LA fire announcement on map to snackbar

### LINKED ISSUE #2381: Change LA Fire announcement position because it's causing layout issues on Mobile
### Overview
We currently have "Due to the LA Fires (Jan 2025), some information may be out-of-date." as an announcement, but it makes the mobile version of the site look weird

image

### Action items

- [ ] Can we have this message responsively move above or below the logo when on mobile? (Desktop can stay the same for now)

Above would be like this: (preferred)
image

Below would be like this, which is kinda awkward
image

- [ ] Change text to: ""Due to the LA Fires, some information may be out-of-date." ~~(Jan 2025)~~

### Related issue

This is where announcements are heading — this is a step towards that.
- #2326

VirginiaWu11:
- Closes #2381 

- moved alert from map page header bar to top right snack bar. 

hey @fancyham, how's this?

Screenshot 2025-03-24 at 6 35 54 PM
Screenshot 2025-03-24 at 6 35 36 PM

### DIFF
--- a/client/src/Routes.jsx
+++ b/client/src/Routes.jsx
@@ -10,7 +10,6 @@ import Toast from "components/UI/Toast";
 import { useWidget } from "./appReducer";
 import useFeatureFlag from "./hooks/useFeatureFlag";
 import SurveySnackbar from "./components/UI/SurveySnackbar";
-import useLocationHook from "hooks/useLocationHook";
 
 const VerificationAdmin = lazy(() =>
   import("components/Admin/VerificationAdmin")
--- a/client/src/Routes.jsx
+++ b/client/src/Routes.jsx
@@ -10,6 +10,8 @@ import Toast from "components/UI/Toast";
 import { useWidget } from "./appReducer";
 import useFeatureFlag from "./hooks/useFeatureFlag";
 import SurveySnackbar from "./components/UI/SurveySnackbar";
+import AlertSurveySnackbar from "./components/UI/AlertSurveySnackbar";
+import useLocationHook from "hooks/useLocationHook";
 
 const VerificationAdmin = lazy(() =>
   import("components/Admin/VerificationAdmin")
@@ -55,6 +57,7 @@ const Suggestion = lazy(() => import("components/FoodSeeker/Suggestion"));
 export default function AppRoutes() {
   const hasUserFeedbackSuveyFeatureFlag = useFeatureFlag("userFeedbackSurvey");
   const location = useLocation();
+  const { isMapPage } = useLocationHook();
 
   return (
     <Suspense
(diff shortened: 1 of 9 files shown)


client/src/components/UI/AlertSurveySnackbar.jsx
```diff
@@ -0,0 +1,58 @@
...
+        >
+          <Stack
+            direction="row"
+            sx={{ alignItems: "center", transform: "translateY(1px)" }}
+          >
+            Due to the LA Fires, some information may be out-of-date.{" "}
+            <IconButton
+              icon="close"
+              sx={{
+                color: alertCloseIcon,
```
VirginiaWu11, line 47: @hanapotski updated color name :D 
didn't know what to name it though lol I also made this one up with a color picker 🤔 

cc @fancyham

fancyham:
Can you make the font color darker and larger?  The font size is looking kinda small on both mobile.  It should probably be same font size as on the pop-up survey dialog.

Please center the snackbar on desktop, too.  We want it to be prominent!

Color-wise: It's feeling low-contrast right now — and though it technically passes WCAG recommendations (checker: https://accessibleweb.com/color-contrast-checker/) but feels hard to see, especially for people with low-vision, and in comparison to the other stuff on-screen (which are 'puncher' and so this notice gets lost)

I think the same dark grey/black fused in the pop-up survey would be good?  It's a notice we want people to read easily.

Also, just thinking out loud here, but I'm wondering if we should have drop shadows for both this and the survey pop-up.  

If it's easy to try both with and without dropshadows and take a screenshot and drop them into the design channel on slack, I'd love that, so we can compare and contrast.

VirginiaWu11:
@fancyham 

Yes. I have just messaged Ingrid on #2382 for colors. Once those are updated, I can go ahead and make the this popup style match the survey popup.

VirginiaWu11:
Hey @fancyham 

How's this?
Updated color & size:

Without drop shadow:
Screenshot 2025-04-07 at 8 52 33 PM
Screenshot 2025-04-07 at 8 52 28 PM


With drop shadow:
Screenshot 2025-04-07 at 8 53 15 PM
Screenshot 2025-04-07 at 8 53 10 PM

fancyham:
Yikes… the alert is blocking the search field and on both mobile and desktop, which is pretty bad.  I think I need to re-think this alert… I'll come up with a few options and run them by you to see what you think.  But for now, let's stop this snackbar/toast direction.  Sorry about that!  On the bright side, seeing things in action and the unexpected repercussions is what building prototypes is all about!

VirginiaWu11:
Hey @fancyham,
I just tried the Hello Bar approach in the video below:

https://github.com/user-attachments/assets/35bd0269-e20c-4780-acea-c02db2bca3fc

Let me know what you think.

client/src/Routes.jsx
```diff
@@ -10,6 +10,7 @@ import Toast from "components/UI/Toast";
 import { useWidget } from "./appReducer";
 import useFeatureFlag from "./hooks/useFeatureFlag";
 import SurveySnackbar from "./components/UI/SurveySnackbar";
+import useLocationHook from "hooks/useLocationHook";
```
hanapotski, line 13: don't think this is being used 👀

VirginiaWu11: you're right!

VirginiaWu11: updated!

hanapotski (approved):

PR #2400, opened 2025-04-06
Fix the react warning on the address dropdown

hanapotski:
We noticed this react warning 👀 

Screenshot 2025-04-06 at 3 25 39 PM


whitneywind (approved):
Looks great!

Issue #2336, opened 2025-01-25
Bug: Some of the organization icons are misplaced on the map


SAUMILDHANKAR:
### Overview
We would like all the meal and pantry icons to be correctly placed on the map. In this issue we will fix some of the misplaced icons.
Following are the details of the two organizations which are obviously misplaced:

1. WISE & Healthy Aging. Address: 1527 4th Street, # 2, Santa Monica 90401
2. Iglesia Fuente De Vida / Vida community Center. Address: 11134 Saticoy Street, Sun Valley 91352

![Image](https://github.com/user-attachments/assets/0fa125b8-5d4d-4bb4-a209-c3beeb50c960)

### Action Items
- [ ] (Optional) write a test to find other icons that might be misplaced. Feel free to create another issue for this if more suitable. 
- [ ] (Optional) explore and implement another additional way to find the latitude and longitude of the organizations and then compare the precision or accuracy of the two methods. Feel free to create another issue for this if more suitable.
- [ ] Discuss with the team about the best way to fix the location of icons on the map and fix it.

### Resources/Instructions
- https://github.com/hackforla/food-oasis/blob/cf037896146f04f0047f5792ed5a21ffcb9ff394/client/src/components/Admin/OrganizationEdit/Identification.jsx#L56

entrotech:
Issue should be closed.

PR #2409, opened 2025-04-14
use viewport range to validate lat & lng input

### LINKED ISSUE #2166: Spellcheck Nit: Postion should be Position
### Overview
- In six variable instances, the word `Position` is misspelled as `Postion`
- Nit picky, but this misspelling could cause headaches down the road if someone instinctually types Position and is confused why their code isn't working. 

### Action Items
- Go to `ResultsMap.js` and `Desktop.js`
- Replace all instances of `listPanelLeftPostion` with `listPanelLeftPosition`
- Replace all instances of `filterPanelLeftPostion` with `filterPanelLeftPosition`
- Replace all instances of `leftPostion` with `leftPosition`

### LINKED ISSUE #2378: Implement Tighter Client-Side Validation of lat/lon
### Overview
We have been having problems where volunteers are not using the geolocation feature on the Organization Edit form to get the lat/lon for listings or mis-typing manually entered lat/lon coordinates, which prevents food seekers from find the listing on the map. This issue is to make lat/lon required and verify that the latitude and longitude values fall within the tenant's lat/lon boundaries.

### Action Items
- [ ] The latitude and longitude ranges are currently stored as constants in the useMapboxGeocoder hook.  Move these constants to the DEFAULT_VIEWPORTS constant object in /clinet/src/helpers/Constants.js and reference it from the useMapboxGeocoder hook.
- [ ] Modify the validationSchema in OrganizationEdit.jsx to apply min and max values to the latitude and longitude value validation.

junjun107:
closes #2378  closes #2166 

viewport bbox bbox: "-118.9517, 33.6988, -117.6462, 34.8233" represents minLng, minLat, maxLng, maxLat

Demo: 
image
This pantry has latitude 34.0112629 and longitude -117.980451

Knowing LA range is 
**minLat** : 33.6988		**maxLat** : 34.8233
**minLng** : -118.9517	        **maxLng** : -117.6462

image
If i lower latitude a bit, alert msg will prompt `latitude must be greater than or equal to 33.6988`
If I lower longitude a bit, alert msg will prompt `longitude must be greater than or equal to
-118.9517|`

### DIFF
--- a/client/src/components/Admin/OrganizationEdit.jsx
+++ b/client/src/components/Admin/OrganizationEdit.jsx
@@ -39,6 +39,8 @@ import {
   LINKEDIN_REGEX,
   PINTEREST_REGEX,
   TWITTER_REGEX,
+  DEFAULT_VIEWPORTS,
+  TENANT_ID,
 } from "../../helpers/Constants";
 
 const phoneRegExp = /^\(\d{3}\) \d{3}-\d{4}$/;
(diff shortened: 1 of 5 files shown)


SAUMILDHANKAR:
Hi @junjun107, this is similar to #2336, my understanding is that if we click on Get Coordinates button, the latitude and longitude will auto populate. So, if they are out of bounds there will be an error, so how will the user resolve the error?

junjun107:
> Hi @junjun107, this is similar to #2336, my understanding is that if we click on Get Coordinates button, the latitude and longitude will auto populate. So, if they are out of bounds there will be an error, so how will the user resolve the error?

Saumil, in organizationedit file, there are two ways to get a stakeholder's cord. 
1. Once address is entered you can use the Get Coordinate button to auto generate the correct lat and lift. When you click set, the numbers will go in to its corresponding input. 
image

2. You can manually enter the lat and lng without using the The stricter validation is intended for this option. 
image

SAUMILDHANKAR:
Thanks June, yes that makes sense. I was thinking about how would the users get the latitude and longitude for an organization. But they can still accidentally enter the lat and long manually so this PR would be very helpful.

PR #2414, opened 2025-04-15
#2251 neighborhood zoom for widget

### LINKED ISSUE #2251: Set a sensible default zoom level when displaying a neighborhood boundary
### Overview
To make the neighborhood council embedded widget look better, set an initial MapBox zoom level such that the neighborhood outlined region fills most of the visible map area.

### Description
[This  wiki page](https://github.com/hackforla/food-oasis/wiki/Embedding-Food-Oasis-in-Your-Web-Site) describes how the Food Oasis app can be embedded in another web site, where it will appear in an iframe. One of the intents for the project is to embed Food Oasis in neighborhood council web sites, and we have implemented a feature that accepts an optional query string parameter containing a neighborhood_id like this:
```https://la.foodoasis.net/widget?neighborhood_id=78```
which will use a shape file from the database to overlay a dashed outline of the neighborhood boundaries on the map.  You can just paste the above url into your browser to see an example.  Though the map is centered on the geographic center of the neighbrohodo area, it just uses an application-default zoom level, so in many cases, the neighborhood occupies just a small part of the map when it is first displayed. 

### Action Items
- [x] Modify the application to automatically zoom such that the neighborhood occupies most of the available map area for any screen size and orientation. You might be able to deduce the lat, lon range from the data already available to the client, or, worst case, we might need to add a column to the database and populate the latititude, longitude ranges in the neighborhood database table.

whitneywind:
Resolves #2251 
This function zooms in on the specified neighborhood when used as a widget. It calculates the bounding box from the region’s GeoJSON, then adjusts the map’s zoom and position to ensure the region fits within the view.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -18,6 +18,7 @@
         "@mui/x-data-grid": "^5.17.20",
         "@mui/x-date-pickers": "^6.8.0",
         "@testing-library/dom": "^10.4.0",
+        "@turf/bbox": "^7.2.0",
         "@types/mapbox-gl": "^3.4.0",
         "@types/node": "^22.9.1",
         "axios": "^1.7.4",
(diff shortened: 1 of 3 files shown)


whitneywind:
A video [demo](https://github.com/hackforla/food-oasis/issues/2251) is in a comment on the issue

hanapotski (approved):

PR #2426, opened 2025-04-28
2129 prevent hitting return while search is loading

### LINKED ISSUE #2129: Bug: Entering an address into the location field and hitting ‘return' quickly breaks search
#### Describe the bug 

On the landing page and on the map screen, if you type in or paste an address and hit return (submit form) before the autocomplete appears, the search fails and user incorrectly gets “Sorry, we don’t have any results for this area”.  Also note that the location field is blank again — ideally we should leave what they typed in the field, or display it in the error.
image

The problem seems to be that the autocomplete doesn’t have a chance to respond.

Also, this problem also seems to happen if someone has a slight misspelling.

The user’s goal is to enter an address, and to get meaningful feedback when there is an error so that the can re-try or make a correction.

- [ ] Add a meaningful error message 'Sorry, we don’t have any results for “[what they typed]” [because you typed too fast.  Please try again./because that address doesn’t exist/because we couldn’t find that address]
- [ ] Consider re-trying the search automatically — clicking “Click here to reset search” seems to get results, even though nothing else changed — or find another solution that lets the system to catch up to the user?
- [ ] Change button to “Click here to retry search” if that’s what it really means.
- [ ] Consider putting what the user typed back into the location field (notice that it’s currently blank after I had entered an address on the previous step)

1. Go to foodoasis.la
2. Paste in an address and hit return immediately, as fast as possible (before autocomplete has a chance to respond)

### What happens

You will see an error message that makes it look like we don’t have listings in that area, even if we actually do.

The accurate results for that address

#### Device configuration
- Device: Mac
- OS version: **Mojave (10.14) which is several years out of date**
- Browser: Safari
- Browser Version: 14.1.2

VirginiaWu11:
# Closes #2129 

https://github.com/user-attachments/assets/e50ff624-101e-46e5-8e5a-3654750d32bd

- fixed debounce so it calls when user stops typing for 300ms
- used fetch loading state to prevent user from hitting enter while loading
- fixed on keyboard up/down arrow navigation & hit enter to go to the correct location

### DIFF
--- a/client/src/components/FoodSeeker/AddressDropDown.jsx
+++ b/client/src/components/FoodSeeker/AddressDropDown.jsx
@@ -23,10 +23,11 @@ export default function AddressDropDown({ autoFocus }) {
     searchCoordinates?.locationName || ""
   );
   const [open, setOpen] = useState(true);
-  const { mapboxResults, fetchMapboxResults } = useMapboxGeocoder();
+  const { mapboxResults, fetchMapboxResults, isLoading } = useMapboxGeocoder();
   const dispatch = useAppDispatch();
   const navigate = useNavigate();
   const { flyTo } = useMapbox();
+  const [highlightedOption, setHighlightedOption] = useState(null);
 
   const handleInputChange = (delta) => {
     if (!delta) return;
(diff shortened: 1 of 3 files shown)


client/src/components/FoodSeeker/AddressDropDown.jsx
```diff
@@ -121,16 +122,21 @@ export default function AddressDropDown({ autoFocus }) {
...
       event.preventDefault();
-      handleAutocompleteOnChange(mapboxResults[0].place_name);
+      const selected = highlightedOption ?? mapboxResults[0].place_name;
+      handleAutocompleteOnChange(selected);
     }
   };
   return (
     <>
       <Autocomplete
+        disableCloseOnSelect
```
SAUMILDHANKAR, line 134: Hi @VirginiaWu11, my expectation was the dropdown wouldn't close if I select an option but it closes automatically. Please share the default behavior for this. Thank you.

SAUMILDHANKAR (approved):

PR #2452, opened 2025-05-24
remove user from sessionStorage to prevent security risk

VirginiaWu11:


client/src/components/PrivateRoute.jsx
junjun107, line 1: Are you keeping the comment here?

hanapotski (approved):

Issue #2366, opened 2025-02-19
Bug: show/hide button for desktop text search results has unexpected behavior


fancyham:
Medium priority — this bug is probably rare right now, but gets user into a state where they can't see their search results.

### Context

In the 2024 redesign, we added the ability to hide the side panel where the text search results appear so that you can see more map.

To hide/show this side panel, we have a 'tab' button that when clicked, hides or shows the side panel.

#### Describe the bugs

There are a few unexpected behaviors and perhaps some that were under-specified, so these are the changes we'd need to make.

If these are very hard to do, we might consider removing the ability to collapse the side panel altogether.  Please research and let us know if that's necessary.

- [ ] The 'hide sidebar' tab is transparent and not visible.  It should be as shown here ([figma](https://www.figma.com/design/9hPmNSsyaH2VMGNz5mffmI/FOLA-Design-%232?node-id=5841-11718&t=CXjfpyIgO86PiQhB-4))
- [ ] When sidebar is collapsed and showing search filters, the 'show sidebar' tab is hidden behind the filters panel and inaccessible.  It should be to the right of the filters panel.
- [ ] When the sidebar is hidden, and user clicks on a pin on the map, the sidebar should automatically show, so that the user can see the details for that pin's associated listing.

#### Videos

**Bug: 'show sidebar' tab becomes inaccessible behind filters panel**
https://github.com/user-attachments/assets/1caf55f2-9d69-48e1-be55-588a3a9d6375

![Screenshot 2025-02-19 at 12 26 46 PM](https://github.com/user-attachments/assets/2c4074f5-c778-4588-a710-7896497ced4a)

**Bug: When pin on map is clicked, the sidebar should reveal if it's currently hidden**
https://github.com/user-attachments/assets/92869697-4d9c-40e1-b99e-b2275bdd729a

**Bug: Tab is transparent**
![Screenshot 2025-02-19 at 12 26 34 PM](https://github.com/user-attachments/assets/a4c0c333-f6f4-428c-a203-68d7409c6705)

#### Device configuration
- Device: Mac
- OS version: Sequoia
- Browser: Safari
- Browser version: 18.3

PR #2453, opened 2025-05-26
fixed show/hide side panel button 

JmJenna:
- Fixed an issue where the hide/show button was not toggling visibility correctly #2366 

Screenshot 2025-05-26 at 9 53 55 AM


client/src/components/FoodSeeker/SearchResults/layouts/Desktop.jsx
```diff
@@ -38,6 +41,8 @@ const DesktopLayout = ({ filters, list, map }) => {
   }));
 
   let leftPosition = isFilterPanelOpen ? "340px" : 0;
+  let listLeftPosition = isFilterPanelOpen ? "-186px" : "-524px";
```
hanapotski, line 44: we can move this inside the function if it's not being used anywhere else 👌

PR #2491, opened 2025-07-01
fix typescript errors in account

VirginiaWu11:


junjun107:
@SAUMILDHANKAR  `announcements: string[]` is not part of User data. So we removed it from server/types/account-types.ts

junjun107 (approved):
LGTM

PR #2492, opened 2025-07-02
make search details shareable via url #2356

### LINKED ISSUE #2356: Bug/Feature: URL should show search terms/address so it can be bookmarked or shared
### Overview

Use case: Person A wants to share a map full of food pantries they've found on the FOLA map.  They text a friend B: "Hey friend, here's a list of pantries near you"… sends URL of search results to their friend… this currently does not work -- B will end up at the generic map but without a location or filters.  It used to work.  See below for why

Let's fix it and make some improvements to support new as well as upcoming features

### Context

People want to be able to share links to search results, but currently they cannot because the only URL we show for the search results is: https://foodoasis.la/organizations

This seems like a regression because we used to show a more complete and sharable URL, such as:
https://foodoasis.la/organizations?latitude=34.09869301323718&longitude=-117.90065104044072

I believe we used to have a similar URL when the user entered an address, too (so not just lat/long)

We currently support neighborhood councils…
https://github.com/hackforla/food-oasis/wiki/Embedding-Food-Oasis-in-Your-Web-Site#showing-a-neighborhood-boundary-on-an-embedded-widget
example URL:
https://foodoasis.la/organizations?latitude=34.09869301323718&longitude=-117.90065104044072&neighborhood_id=78

### How is it now?
Currently, a search result is not-bookmarkable, blocking our users from sharing and saving their progress, and from sharing search results with their family or friends. 
 
### Proposed bug fix / new feature

- [ ] Embed the user's (search address or lat/long, as apporpriate), **search terms (i.e. pantry, meal program, other filters such as the upcoming text search or org name search)**, and neighborhood council into the URL
- [ ] Likewise, put all selected filters and related arguments into the URL.  This might include things like a neighborhood-outline-overlay.
- [ ] Put the updated URL in the browser address bar anytime one of those change so that it can be bookmarked or shared with a friend

Think of this as creating an API that other apps (or users) can use to access our data

whitneywind:
Resolves #2356 

- This PR adds the user's search info to the URL so that the specifics can be shared to saved for later. 
- The details included in the URL are the search terms (if applicable), pantry/meal filter, open time, and location (latitude, longitude, and zoom amount)

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.jsx
@@ -1,41 +1,18 @@
 import { Box, Stack, Typography } from "@mui/material";
 import Grid2 from "@mui/material/Unstable_Grid2";
 import AddressDropDown from "components/FoodSeeker/AddressDropDown";
-import {
-  FOOD_PANTRY_CATEGORY_ID,
-  MEAL_PROGRAM_CATEGORY_ID,
-} from "constants/stakeholder";
 import PropTypes from "prop-types";
-import { useCallback } from "react";
-import * as analytics from "services/analytics";
-import CategoryButton from "./CategoryButton";
 import SwitchViewsButton from "./SwitchViewsButton";
-import useFeatureFlag from "hooks/useFeatureFlag";
 import { TENANT_CONFIG } from "../../../../helpers/Constants";
 import Geolocate from "./Geolocate";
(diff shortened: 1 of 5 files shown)


SAUMILDHANKAR (approved):
Thank you @whitneywind, this is looking very good.

I am not sure if I tested right but when I used ```https://curly-guide-4j7vp49j99cx7r-3000.app.github.dev/organizations?pantry=1&meal=1&lat=33.8143058&lng=-118.4485516&zoom=11.8&openRadio=Open+Now&name=mart```
the pin drop location changed:
before url
after url

PR #2536, opened 2025-08-26
2527 reorder search dropdown to show map address before organizations

### LINKED ISSUE #2527: On org name/address each field: put addresses above org names in autocomplete
### Overview
The combo org name / address search is super powerful but we can improve the order of the results by putting address results before org name results.

### How it is now

@ihorng Discovered this unexpected behavior:

People searching are most commonly entering an address, and less often, searching for org names. 

However, our autocomplete currently has org names at the top of the autocomplete, making it hard to search for, say, "Hollywood" or "90015" and hitting return to get an expected result.

Example:

Seach for "hollywood"

Org names are about 80% of the results

Image

And Hollywood the neighborhood/city is almost at the bottom (note the scroll bar)
Image

### Action items

**So let's swap the order -- addresses before org names.**  

That way neighborhoods appear when someone types 90015 or "hollywood"

- [x] Change order in autocomplete so that  address results appear before org name results
- [x] (this may be hard) If possible, add a divider in the autocomplete drop-down: "Addresses" above the addresses, "Organizations" above the organizations.
- [ ] Stretch goal: If possible, for orgs that have the same name, show the address underneath (or at least visible in dropdown) so that user can disambiguate

VirginiaWu11:
- Closes #2527

### DIFF
--- a/client/src/components/FoodSeeker/AddressDropDown.jsx
+++ b/client/src/components/FoodSeeker/AddressDropDown.jsx
@@ -173,7 +173,7 @@ export default function AddressDropDown({ autoFocus }) {
     }
   };
 
-  const combinedOptions = [...stakeholderOptions, ...mapboxOptions];
+  const combinedOptions = [...mapboxOptions, ...stakeholderOptions];
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


SAUMILDHANKAR (approved):
Thanks for this 5 star implementation.

PS: 1) Should we make a new issue for the stretch goal since this issue will closed upon merging.
2) Not related to this issue but I noticed when we select an organization, instead of that org icon getting larger, we add a new icon, this was confusing for me and I don't remember if that is the design and can't find anything related to reorder on figma. Here is my recording:

https://github.com/user-attachments/assets/82462395-6bb7-4ac5-a214-372fe1d414e4

fancyham:
Nice!

PR #2537, opened 2025-08-26
updated map pins #2403

### LINKED ISSUE #2403: Bug: Update the map pins to be consistent in style and accurately reflect the correct locations
**Bug description:**
Map pins are inconsistent in appearance and in some cases, inaccurately placed on the map. This affects all users trying to find specific locations using the map.

**Who it affects:**
All users, particularly those who rely on the map to locate services (ex: food pantries, meal programs, etc)

**How it affects users:**
Users may become confused or frustrated if a pin doesn’t accurately represent a location, leading to potential misdirection or wasted time. Inconsistent pin styles can also reduce trust in the interface and make the map harder to scan quickly.

**User’s goal:**
To quickly and confidently find the correct location of a service or resource using the map.

**Screenshot of issue:**

1. The size of the hybrid pins are smaller and missing the shadow 

Image

####

2. When a pin is selected, the origin of the pin shifts

Image

####

Image

####

**What's the expected result?**
All map pins should be consistently styled and accurately placed to reflect the true locations of services. Users should be able to quickly understand what each pin represents and trust that tapping on a pin provides accurate information.

**Solution:**
- Currently, the hybrid pin on the website is smaller and does not have a shadow. Change the hybrid pin to match the size as the food pantry and meal program pins, which should be 36x58 px. Attached is the link to the figma file of the detailed specs and instruction: https://www.figma.com/design/9hPmNSsyaH2VMGNz5mffmI/FOLA-Design--2?node-id=9240-9306&t=YZhbDiNpIJJr8fUK-4
- Export the (hybrid) pin with no shadow from the figma file
- Add the shadow using CSS
- The origin of all map pin SVGs should be set at the pointy tip of the icons, so they don’t shift position when selected or deselected during interaction on the map

Image

whitneywind:
- resolves #2403 
- added the 6 map icons from the figma (3 selected and 3 unselected) and converted them to react svg components
- replaced the mapmarker file to use the updated icons
- the map pin tips now stay in the same location when selected or unselected


whitneywind:
https://github.com/user-attachments/assets/0a32ce8a-7cc8-4aed-81b3-a98c00780399

SAUMILDHANKAR (commented):
Hi @whitneywind thank you for working on this issue. I noticed that the split icon still moves just a tiny bit when clicked, please see the recording below:

https://github.com/user-attachments/assets/d5e94452-dea8-4300-a6ef-4401645fcfde

- Also, I noticed a mismatch of icon type in the side-panel and the map, for example: organization ```Onegeneration``` should be a split icon but on the map it is coming as a meal icon. Maybe it is just my end.
- And I checked that the files ```client/src/icons/MealIcon.jsx```, ```client/src/icons/PantryIcon.jsx``` and ```client/src/icons/SplitPantryMealIcon.jsx``` won't be used anywhere after your changes so maybe we can delete them?

Let me know your thoughts, I will double check.

SAUMILDHANKAR:
Thank you Whitney. I verified the icons are now in order. It is amazing how you noticed that the icon frames are slightly off center. Nice work. Let us discuss this with the design team if they want something else.

PR #2547, opened 2025-09-14
feat filter panel custom hour any

### LINKED ISSUE #2385: Date and time filter should not require both fields be filled out, and should allow "any"
#### Overview

We want to make the day/time selector more useful by allowing "any" as options for day and / or time.

### How it is now

Right now, if you use this filter, you must choose **both** a day and time before results update.
image

But some users are looking for 
- Places open anytime on Monday
- Places open at 7pm (after work) on any day

So let's help them by:

### Change requested

- [ ] **Update results as soon as a day or time is selected.**  If a day or time filter is not-yet-selected, then that should be counted as a wildcard any/all/*
- [ ] **Add 'Any' as an option to the bottom day and time dropdown menus and filter accordingly**  Currently, there's no way to select "any" once you've selected a day or time, so let's add that ability.

For example, this should filter results to places open Sunday (at any time)
image

**Stretch goal, if you like a challenge:**

- [ ] Add "Weekday" and "Weekend" to the bottom of the Day filter dropdown menu, above "Any".

1. Go to la.foodoasis.net
2. Enter a zip code
3. Select 'more filters'
4. Open Time > Day : Select 'SUN'

What happens:
- Results don't update.  Must also select a time.

What's expected:
- Results update to show places open on Sunday at any time.

This same process applies to the time selector, too.

junjun107:
Closes #2385 

## WHAT
Open Time filter logic was updated to support more flexible custom day/time selections

* Day + Time → keeps existing behavior (filters by exact day and time).

* Day + Any → new filter logic, If user picks SUN + "Any", you’ll get all Sunday results.

* Time + Any day → new filter logic , If user don't pick a day + "11AM", you’ll get all results for 11AM

*  Both empty → shows all results (no filtering).

## What this PR Changed / Updated

Refactored useOrganizationBests filtering to handle four clear cases:

1. Day + Time
2. Day only (any time)
3. Time only (any day)
4. No filter

Replaced old "SUN" hack in time-only filtering with a check across all days.
Added "Any" to the time dropdown 

## Media

Screenshot 2025-09-13 at 9 03 16 PM
Screenshot 2025-09-13 at 9 03 54 PM
Screenshot 2025-09-13 at 9 05 01 PM

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.jsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.jsx
@@ -345,6 +345,7 @@ const OPEN_TIME_DROPDOWNS = {
   time: {
     placeholder: "Open Time",
     labels: [
+      "Any",
       "12:00AM",
       "01:00AM",
       "02:00AM",
--- a/client/src/helpers/index.js
+++ b/client/src/helpers/index.js
@@ -218,10 +218,10 @@ export const computeDistances = (userLatitude, userLongitude, stakeholders) => {
  * Example output: "2023-12-24T06:00:00.003Z"
  */
 export const getNextDateForDay = (dayInput, timeInput, targetTimezone) => {
-  if (timeInput === "Any") {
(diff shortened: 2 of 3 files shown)


VirginiaWu11 (approved):
LGTM

PR #2558, opened 2025-09-30
2504 filter announcements

### LINKED ISSUE #2504: Add filter feature to sort announcements
### Overview
Currently, the announcements get added based on when they were created. We would like them to be custom sorted by the admin for instance there should be an option to sort by severity or by number of days when the announcement was created. In this issue, we will add a filter to achieve this functionality.

### Action Items
- [ ] Write the pseudo code to achieve above functionality
- [ ] Check with team about severity levels
- [ ] Write breaking test first using TDD (test driven development)
- [ ] Test it in the local environment
- [ ] Once the functionality is achieved check if code can be refactored

### Resources/Instructions
- Related file: https://github.com/hackforla/food-oasis/blob/develop/client/src/components/Admin/Announcements.jsx

Giolii:
Fixes  #2504 

Add filter functionality to admin Announcements


client/tests/announcements.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+          description: "We have updated our system.",
+          created_at: "2025-07-18T15:45:00.000Z",
+          severity: "warning",
+          is_enabled: true,
+        },
+      ],
+    });
+
+    await helpers.mockLogin();
+    await page.goto("/admin/announcements");
```
hanapotski, line 29: I just realized the login state is not persistent in the tests. We need to fix that 🫣 but for now, you can do this instead of navigating to another page 👌 
```
    await page.getByTestId("menu-button").click();
    await page.getByRole("link", { name: "Announcements" }).click();
```

PR #2562, opened 2025-10-04
refactor AccountAutocomplete and AssignDialog components TypeScript

### LINKED ISSUE #2543: Convert AssignDialog to typescript
### Overview
We need to convert the AssignDialog react component, to make it maintainable by future developers.

### Action Items
- [x] Convert AssignDialog to typescript

### Resources/Instructions
n/a

junjun107:
closes #2543


client/src/components/Admin/AccountAutocomplete.tsx
```diff
@@ -1,15 +1,30 @@
 import { Autocomplete, TextField } from "@mui/material";
 import { useAccounts } from "../../hooks/useAccounts";
 
+interface Account {
+  id: string | number;
+  firstName: string;
+  lastName: string;
+  email: string;
+}
```
hanapotski, line 9: should we move this to a separate file so we can use it in the service file?

client/src/components/Admin/AccountAutocomplete.tsx
```diff
@@ -1,15 +1,30 @@
...
+interface AccountAutocompleteProps {
+  accountId: string | number | null;
+  setAccount?: (account: Account | null) => void;
+  setAccountId?: (id: string | number | null) => void;
+  label?: string;
+  name?: string;
```
hanapotski, line 15: this isn't being used 👀

client/src/components/Admin/AccountAutocomplete.tsx
```diff
@@ -1,15 +1,30 @@
...
   setAccount,
   setAccountId,
   label,
-}) {
+  name,
+}: AccountAutocompleteProps) {
   const { data: accounts } = useAccounts();
 
-  const handleChange = (event, value) => {
+  const handleChange = (_event: unknown, value: Account | null) => {
```
hanapotski, line 27: I think this event is `React.SyntheticEvent` but if we're not using it, we prob shouldn't pass it in 🤔

VirginiaWu11 (approved):

PR #2594, opened 2025-11-09
Add a Playwright auth state JSON file for reuse  

junjun107:
## What
Playwright tests were re-running login for every test. This PR saves the logged in state and creating a HTML test reports under the tests folder.

## Changes

- Added client/playwright/auth.setup.ts to log in once and save the auth state to .auth-state/testuser.json.
- Updated playwright.config.ts to:
   - Reuse saved login state for all E2E tests.
   - Keep auth.spec.ts logged out.
   - Generate Playwright reports in tests/playwright-report.
- Added tests/playwright report to .gitignore.

## Steps to test
1. Generate the saved auth sate: 
`npx playwright test playwright/auth.setup.ts`
2. Run all tests: 
`npx playwright test  `
3. View the HTML report: Run ` npx playwright show-report tests/playwright-report` or click on it 
image

## Example of HTML test report 
image

### DIFF
--- a/client/tests/announcements.spec.ts
+++ b/client/tests/announcements.spec.ts
@@ -3,10 +3,12 @@ import test from "./helpers/test";
 import { expect } from "@playwright/test";
 
 test.describe("Announcements", () => {
-  test("Should display announcements page with table", async ({
-    page,
-    helpers,
-  }) => {
+  test.beforeEach(async ({ page, helpers }) => {
+    await mockRequests(page);
+    await helpers.mockLogin();
+  });
+
+  test("Should display announcements page with table", async ({ page }) => {
     await mockRequests(page, {
       announcements: [
         {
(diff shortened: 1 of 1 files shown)


Giolii:
Tested locally and the changes work well

But I have still problems, I went back to `announcements.spec.ts` and changed the fix Hana suggested to the old way who was failing in PR #2558:

Screenshot 2025-11-09 at 12 37 13 AM

It still doesn't pass if I use `page.goto`. Am I missing something?

Screenshot 2025-11-09 at 12 45 09 AM

junjun107:
Ahhh I finally understand what the issue is. You are trying to use `page.goto()` to navigate to /admin/announcements. When you do that, Playwright directly loads that URL from your dev server.
But the server doesn’t automatically route all paths back to your React app’s index.html, the app never starts, so nothing appears. (Common problem of SPA)
But when you click through the menu like a user, React Router handles the route change in the browser, fetches data, and your mocks work as expected. 

TLDR: The mocks don’t cover server-side fetches during a hard goto. 

I suggest writing tests that mimic real user actions through the UI.

PR #2672, opened 2026-02-03
Fix scroll position not resetting between routes

### LINKED ISSUE #2662: Bug:  Scroll position not resetting when switching sidebar tabs
#### Describe the bug 

When switching between tabs in the sidebar, the scroll position persists from the previous tab instead of resetting. This causes users to land in the middle or bottom of a new tab's content rather than at the top.

Open the app
Navigate to a tab in the sidebar (e.g., About us)
Scroll down within that tab's content
Click on a different tab in the sidebar (e.g., FAQs)
Observe the scroll position

- When switching to a new tab, the scroll position should reset to the top of the content area, allowing users to view the new tab's content from the beginning.

- The scroll position from the previous tab carries over to the new tab, leaving users at the old position in the content area.

Image

Giolii:
#### Summary

  Fixes scroll position not resetting when navigating between static pages (About, Donate, FAQs, Contact, Suggest New Listing).

  #### Problem

When users scroll down on one page and navigate to another page via the sidebar menu, the new page would maintain the previous scroll position instead of starting at the top.

  This happened because the scroll was occurring on the Grid container with fixed height (100vh) rather than the browser window itself.

  #### Solution

  - Created a ScrollToTop component that uses React Router's useLocation hook to detect route changes
  - Added id="app-container" to the main Grid container for targeted scrolling
  - Component scrolls both the window and the app container to top on every route change

  #### Files Changed

  - client/src/components/ScrollToTop.jsx (new)
  - client/src/Routes.jsx

 Fixes #2662

### DIFF
--- a/client/src/Routes.jsx
+++ b/client/src/Routes.jsx
@@ -11,6 +11,7 @@ import { useWidget } from "./appReducer";
 import useFeatureFlag from "./hooks/useFeatureFlag";
 import SurveySnackbar from "./components/UI/SurveySnackbar";
 import AnnouncementSnackbar from "components/UI/AnnouncementSnackbar";
+import ScrollToTop from "./components/ScrollToTop";
 
 const VerificationAdmin = lazy(() =>
   import("components/Admin/VerificationAdmin")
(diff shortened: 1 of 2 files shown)


junjun107 (approved):
LGTM

PR #2710, opened 2026-03-21
Add server side validation for stakeholder

### LINKED ISSUE #1786: Add server side validation for Stakeholder

Giolii:
## Summary
Add server-side validation for stakeholder `POST` and `PUT` requests.

  ## Problem
Stakeholder create and update requests were not being validated on the server.

  Because of that:
  - invalid `POST` requests could still create stakeholders
  - invalid `PUT` requests could still update stakeholders
  - bad values could reach the database and fail later with a `500` instead of being rejected with `400`
  
A bad actor can bypass frontend validation with tools like Postman and submit invalid payloads.

  ## Solution
  Add validation schemas for stakeholder `POST` and `PUT`, then apply them in the stakeholder router before the controller runs.

  On the frontend:
  - convert `latitude` and `longitude` to numbers before submit
  - normalize `tags` from `null` to `[]` before submit

| Before | After |
|--------|-------|
| invalid post missing name | invalid post 400 |
| invalid put missing name | invalid put 400 no name |
| Invalid PUT with non-numeric coordinates | invalid put 400 string coordinates |

  ## Changes

  - added stakeholder `POST` and `PUT` schemas
  - added `requestValidationMiddleware` to stakeholder `POST` and `PUT`
  - updated frontend stakeholder submit flow to send numeric `latitude` / `longitude`
  - normalized `tags` to `[]` on submit instead of `null` 
  - added backend `POST` and `PUT` tests

  ## Test

run npm test on "client"
run npm test on "server"
Test backend manually: Follow screenshots in "Solution"  to test endpoints
Test frontend manually: Check the screenshots at the end to test the UI.

UI Buttons who send PUT and POST requests:
- Save Progress (POST when crating a new stakeholder, PUT when editing an existing stakeholder)
- Needs Verification (PUT)
- (Re-)Assign (PUT)
- Request Changes (PUT)
- Approve (PUT)
- Hand Off (PUT)
- Submit For Review (PUT)

### Data Entry User
Screenshot 2026-03-20 at 3 28 30 PM

### Admin user
Screenshot 2026-03-20 at 4 00 39 PM

Fixes #1786

### DIFF
--- a/server/app/validation-schema/stakeholder-schema.ts
+++ b/server/app/validation-schema/stakeholder-schema.ts
@@ -1,0 +1,375 @@
+import { JSONSchemaType } from "ajv";
+
+type StakeholderHour = {
+  close: string;
+  dayOfWeek: string;
+  open: string;
+  weekOfMonth: number;
+};
+
+type StakeholderPostRequest = {
+  address1: string;
+  address2: string;
+  adminContactEmail: string;
+  adminContactName: string;
+  adminContactPhone: string;
+  adminNotes: string;
+  allowWalkins: boolean;
+  categoryNotes: string;
+  city: string;
+  confirmedAddress: boolean;
+  confirmedCategories: boolean;
+  confirmedEmail: boolean;
+  confirmedFoodTypes: boolean;
+  confirmedHours: boolean;
+  confirmedName: boolean;
+  confirmedPhone: boolean;
+  covidNotes: string;
+  description: string;
+  donationAcceptFrozen: boolean;
+  donationAcceptPerishable: boolean;
+  donationAcceptRefrigerated: boolean;
+  donationContactEmail: string;
+  donationContactName: string;
+  donationContactPhone: string;
+  donationDeliveryInstructions: string;
+  donationNotes: string;
+  donationPickup: boolean;
+  donationSchedule: string;
+  eligibilityNotes: string;
+  email: string;
+  facebook: string;
+  foodBakery: boolean;
+  foodDairy: boolean;
+  foodDryGoods: boolean;
+  foodMeat: boolean;
+  foodPrepared: boolean;
+  foodProduce: boolean;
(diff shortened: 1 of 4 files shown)


SAUMILDHANKAR (approved):
Very well-done PR. Thank you.

server/app/validation-schema/stakeholder-schema.ts
```diff
@@ -0,0 +1,375 @@
...
+      open: {
+        type: "string",
+        minLength: 1,
+      },
+      close: {
+        type: "string",
+        minLength: 1,
+      },
+    },
+    additionalProperties: true,
```
SAUMILDHANKAR, line 106: I understand we are leaving this true so that app doesn't break in case there are properties added on the front-end.

server/app/validation-schema/stakeholder-schema.ts
```diff
@@ -0,0 +1,375 @@
...
+  state: {
+    type: "string",
+    minLength: 1,
+  },
+  zip: {
+    type: "string",
+    minLength: 1,
+  },
+  latitude: {
+    type: "number",
```
SAUMILDHANKAR, line 136: Nit-picky, we may add minLength here?

server/app/validation-schema/stakeholder-schema.ts
Giolii: Yes! Mainly compatibility with the current UI

server/app/validation-schema/stakeholder-schema.ts
Giolii: Good observation. Since it's a number, I can validate it with a minimum and maximum, but that would hardcode the LA range

SAUMILDHANKAR: oops there is no minLength on a number

PR #2713, opened 2026-03-22
Restrict data entry users to assigned stakeholders on server-side

### LINKED ISSUE #2711: Bug: Every "Data entry" user can edit the stakeholder informations
#### Describe the bug 

The current logic lets every data entry user edit any stakeholder, if they type the corresponding url for the Organization Edit form into the address bar.

1. Log in as a Data Entry user
2. Type the following into the browser address bar: `https://devla.foodoasis.net/admin/organizationedit/4210`
3. You will be able to view and edit the entry, even though the listing has not been assigned to you.

- We probably want to implement some special logic that prevents you from opening this form for listings that have not been explicitly assigned to you as a Data Entry user.

#### Goal

Implement a way to check if the user has been assigned to a stakeholder. If not, redirect user  to some other page (maybe their Dashboard)?  The Web  API endpoint for saving changes should also only proceed if the user is either a data entry user that has been assigned to the listing, or is an admin, as a fallback to protect the app from allowing malicious users from making unauthorized HTTP requests.

Giolii:
## Summary
Add server-side restriction for stakeholder `GET` and `PUT` requests so backend access matches the UI restrictions for Data Entry users.

  ## Problem
Any Data Entry user can read and update stakeholder records through postman or the terminal, bypassing the UI restrictions.  

  ## Solution
  Add server-side validation to verify that a Data Entry user making a GET or PUT request is assigned to the stakeholder they are trying to read or modify.

| Before | After |
|--------|-------|
|  2711 before |  2711 after |
|  2711 before 1 | 2711 after 1  |

  ## Changes

- stakeholder-controller.ts
      - Add a server-side ownership check for Data Entry users
      - Block GET and PUT when the authenticated Data Entry user is not assigned to the stakeholder
      - requestAssignment now uses the authenticated user id from the JWT instead of client-provided (request.body) loginId
- stakeholder-service.ts
      - Add isStakeholderAssignedToUser to verify whether a stakeholder is assigned to a specific user
- express.d.ts
      - Add sub to the request user type so the controller can read user roles from the JWT payload
- stakeholder-controller.test.ts
      - Add tests to verify:
          - unassigned stakeholders cannot be read by Data Entry users
          - assigned stakeholders can be read by Data Entry users
          - unassigned stakeholders cannot be updated by Data Entry users
          - admin users are not blocked by this restriction

  ## Test

npm test on `server`
npm run typecheck on `server`
npm test on `client`
npm run build 
Test backend manually: Follow screenshots  to test endpoints
Test frontend manually: Login with data entry user, request an assignment, try to change the ID in the URL to access another stakeholder(should show 403 on console and an empty form).

Fixes #2711

### DIFF
--- a/server/app/controllers/stakeholder-controller.ts
+++ b/server/app/controllers/stakeholder-controller.ts
@@ -2,6 +2,10 @@ import stakeholderService from "../services/stakeholder-service";
 import { Readable } from "stream";
 import { RequestHandler } from "express";
 import {
+  getAuthenticatedUserId,
+  hasStakeholderAccess,
+} from "../helpers/stakeholder-access";
+import {
   StakeholderSearchParams,
   Stakeholder,
   ClaimParams,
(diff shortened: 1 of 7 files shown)


Giolii:
As suggested in the meeting I am going to:

- [x] Delete the redundant if statement

- [x] Change the name of the function from  `ensureStakeholderAccess` to  `assertStakeholderAccess`

- [x] Move `ensureStakeholderAccess` to a file in the helper folder so it can be exported and reused

server/app/controllers/stakeholder-controller.ts
```diff
@@ -61,6 +65,16 @@ const getById: RequestHandler<
...
+    const stakeholderId = Number(req.params.id);
+    const userId = getAuthenticatedUserId(req);
+    if (!userId) {
+      res.sendStatus(401);
+      return;
+    }
+    if (!(await hasStakeholderAccess(req.user, stakeholderId))) {
+      res.sendStatus(403);
+      return;
+    }
```
hanapotski, line 77: I was thinking we can maybe do something like this 🤔 

```
 await assertHasStakeholderAccess(req.user, stakeholderId);
```

so inside the function, we'll just throw an error?

```
type StakeholderAccessError = Error & {
  statusCode: number;
};
export const assertHasStakeholderAccess = async (
  user: StakeholderAccessUser | undefined,
  stakeholderId: number
) => {
  const hasAccess = await hasStakeholderAccess(user, stakeholderId);
  if (hasAccess) return;

  const error = new Error("Forbidden") as StakeholderAccessError;
  error.statusCode = 403;
  throw error;
};
```

Giolii: got it! can you check now?

hanapotski (approved):

PR #2773, opened 2026-05-11
Fix address bar behavior on fast typing

### LINKED ISSUE #2598: Bug: fix the behavior of the address bar so that it does not go to the wrong address when using enter key
### Overview
We need to fix the behavior of the address bar so that it works properly when using the return button on keyboard

#### Describe the bug 

Typing in a zip code and pressing return quickly will often go to the wrong address.

Seems to be an issue with the autocomplete not completing in time, and sometimes even using old autocomplete data?

Please see video:

1. Go to la.foodoasis.net
2. Type in '90015' + 'return' quickly
4. The address field sometimes gets filled in with the first organization in the autocomplete.  Other times, it gets filled in with a previously-used address?  See video for more examples.

If I type in '90015' and hit return, I expect to go to "Los Angeles, California, 90015, USA" per the first auto-complete result.

The incorrect address

Video

https://github.com/user-attachments/assets/23e00b7c-b1d3-45c0-97a2-5a687370d7b2

#### Device configuration
- Device: MacBook
- OS version: Sequoia 15.6.1
- Browser: Safari
- Browser version: 18.6

Giolii:
## Summary

  Fix the search dropdown race condition where pressing Enter quickly after typing a zip code could select a stale organization instead of the correct address.

  ## Problem

  The dropdown mixed two different states:

  - the current input text
  - the last rendered autocomplete results/highlighted option

  When a user typed a zip code like 90011 and pressed Enter quickly, the debounced address lookup could still be behind. In that window, the component could use a stale highlighted option from the previous result, which sometimes caused it to navigate to the first organization instead of the first address.

  ## Solution

  Move result-freshness handling into the geocoder hook and make the dropdown use that API.

  Changes made:

  - useMapboxGeocoder now tracks which query the current Mapbox results belong to with resultsQuery
  - the hook ignores stale async responses so older searches cannot overwrite newer ones
  - the hook now exposes:
      - searchMapboxResults for normal debounced typing
      - ensureMapboxResults for resolving results for the current input on Enter
  - AddressDropDown now clears any stale highlighted option when the input changes
  - on Enter, the dropdown only trusts the highlighted Mapbox address if the rendered results already match the current input, otherwise it asks the hook for current results and selects the first address

  ## Test

npx playwright test

  Manual verification:

  - type a zip code such as 90011
  - press Enter normally
  - press Enter very quickly after typing

  Expected result in both cases:

  - navigation uses the first address result for the current zip code
  - it no longer falls through to the first organization due to stale highlighted results
  
  Closes #2598

### DIFF
--- a/client/src/components/FoodSeeker/AddressDropDown.jsx
+++ b/client/src/components/FoodSeeker/AddressDropDown.jsx
@@ -28,7 +28,13 @@ export default function AddressDropDown({ autoFocus }) {
     searchCoordinates?.locationName || ""
   );
   const [open, setOpen] = useState(false);
-  const { mapboxResults, fetchMapboxResults, isLoading } = useMapboxGeocoder();
+  const {
+    mapboxResults,
+    searchMapboxResults,
+    ensureMapboxResults,
+    isLoading,
+    resultsQuery,
+  } = useMapboxGeocoder();
   const dispatch = useAppDispatch();
   const navigate = useNavigate();
   const { flyTo } = useMapbox();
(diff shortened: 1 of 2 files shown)


junjun107 (approved):
A bit overkill imo but the race guard is correct and it fixe the box! 
Ship it.

PR #2792, opened 2026-06-02
update twitter to x in stakeholder

### LINKED ISSUE #2172: Update twitter URLs
### Overview
We would like all the URLs in our application to be up to date. In this issue, we ill search for all the "twitter.com" URLs and replace them with "x.com".

### Action Items
- [ ] Search the codebase for twitter.com URL, make sure to check if the URL is generated dynamically using concatenation.
- [ ] Update the URL to x.com
- [ ] Test the links to make sure the link redirects correctly.

### Resources/Instructions
- https://github.com/hackforla/food-oasis/blob/df35240ad3ac35929f2ba749fb91654b94d1820d/client/src/components/StaticPagesHI/Content/pageAbout.js#L160

whitneywind:
Resolves #2172 

- Updates Twitter URLs in the stakeholder table to use x.com by replacing. Includes a rollback migration to revert the change if needed.

### DIFF
--- a/server/migrations/1776360483357_update-twitter-urls-2172.js
+++ b/server/migrations/1776360483357_update-twitter-urls-2172.js
@@ -1,0 +1,19 @@
+/* eslint-disable camelcase */
+
+exports.shorthands = undefined;
+
+exports.up = (pgm) => {
+  pgm.sql(`
+    UPDATE stakeholder
+    SET twitter = REPLACE(twitter, 'twitter.com', 'x.com')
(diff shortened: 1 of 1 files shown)


entrotech (approved):
Looks good to me.  It won't fix urls where they don't include the domain name at all, but it does what the issue asked for.

whitneywind:
Thanks for reviewing! Right, this change doesn't impact the bare handles like @username, plain usernames without a domain, or any of the "N/A" strings that exist in the column.

PR #2813, opened 2026-06-16
Widen address bar #2596

### LINKED ISSUE #2596: Make the address bar wider because we have space available
### Summary
The address bar is currently cramped and also reduces usability.  This issue fixes that.

This should be relatively easy to program and will make the address bar more useful, both for displaying the current address, but also for showing more address autocomplete text without truncation.

### Overview

Make the address/org name search bar bigger so user can see more of what was entered.  Use up more of the available whitespace on the row.

Currently, the bar is using up less space than is available.

It may be because there are two hidden 'food pantry' and 'meal program' buttons that are now hidden.   Those two buttons can be removed now as they've been replaced by the filter buttons that float above the map.

**Wide mobile view (think iPad) on left, desktop view on right**
Image

### Action Items

- [ ] Make it more like below where the address bar uses more of the available space.  It should have some reasonable maximum width so that it doesn't look totally weird on a wide screen. The address bar can use up all the available space on narrower screens, but be aware we may add other elements on that row in the future:
Image

- [ ] Also, try to fix the weird text wrap of the "Locate free food in Los Angeles" when the screen is narrow

whitneywind:
- Resolves #2596 
- Widens the address bar to fill the horizontal space, showing the whole address when screen size allows for desktop and mobile 
- Removes old feature flag and references related to map filters

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
@@ -29,7 +29,7 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
     >
       <Grid2
         display={{ xs: "none", sm: "block" }}
-        sm={4}
+        sm="auto"
         sx={{
           paddingLeft: "1rem",
         }}
@@ -50,7 +50,7 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
       <Grid2
         container
         xs={12}
-        sm={8}
+        sm={true}
         justifyContent="center"
         alignItems="center"
       >
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
@@ -5,7 +5,6 @@ import AddressDropDown from "components/FoodSeeker/AddressDropDown";
 import SwitchViewsButton from "./SwitchViewsButton";
 import { TENANT_CONFIG } from "../../../../helpers/Constants";
 import Geolocate from "./Geolocate";
-import useBreakpoints from "hooks/useBreakpoints";
 
 interface ResultsFiltersProps {
   showList: boolean;
@@ -17,7 +16,6 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
   toggleShowList,
 }) => {
   const { taglineText } = TENANT_CONFIG;
-  const { isMobile } = useBreakpoints();
 
   return (
     <Grid2
@@ -31,7 +29,7 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
     >
       <Grid2
         display={{ xs: "none", sm: "block" }}
-        sm={4}
+        sm="auto"
         sx={{
           paddingLeft: "1rem",
         }}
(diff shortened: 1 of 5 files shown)


whitneywind:
https://github.com/user-attachments/assets/163ac48f-d919-421f-99ef-ed9cb4c98f51

client/src/components/FoodSeeker/SearchResults/layouts/Mobile.tsx
```diff
@@ -49,17 +47,15 @@ const MobileLayout: FC<MobileLayoutProps> = ({ filters, map, list, showList }) =
     }
 
     setPosition({ x: 0, y: newY * (window.innerHeight / 100) });
-  }, [showList, filterPanelOpen, hasAdvancedFilterFeatureFlag]);
+  }, [showList, filterPanelOpen]);
```
SAUMILDHANKAR, line 50: Sorry, I am not aware if we are removing the feature flag, so wanted to confirm. Also, I notice the draggable component in your branch and dev branch behave differently than the prod env, so I am not sure about the requirement.

client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
```diff
@@ -52,45 +50,37 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
       <Grid2
         container
         xs={12}
-        sm={8}
+        sm={true}
```
SAUMILDHANKAR, line 53: Copilot suggesting ```sm={true}```
and others like
```
direction={{ xs: "column-reverse", sm: "row", }}
```
may need to wait till we upgrade to MUIv7.
Quoting copilot:
```In MUI v5, the direction prop only accepts a single string ("row", "column", "row-reverse", "column-reverse").

Responsive objects ({ xs: ..., sm: ... }) were not supported for direction in v5.
```

client/src/components/FoodSeeker/SearchResults/layouts/Mobile.tsx
whitneywind: Yes, there is a comment on the issue requesting that the feature flag be removed, and it was discussed briefly in last week's dev meeting!

client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
whitneywind: They appear to work well on our current version of MUI. The `sm={true}` is what makes the address bar fill up the appropriate amount of remaining space. 
The other code like`direction={{ xs: "column-reverse", sm: "row", }}` is how these components are currently written; this PR doesn't change those.

client/src/components/FoodSeeker/SearchResults/layouts/Mobile.tsx
SAUMILDHANKAR: My bad, but thank you so much for the clarification. Although I am having a hard time testing if this feature flag is removed, I can't find the .env.local file. Recalling from earlier, it supposed to include the name of feature flag as environment variable. Please allow me time as I work through this or if you have the visual for the feature flag removed, we can merge this asap. Thank you. 👍

whitneywind: Thank you, Saumil! You can see the feature flag as an option on both the dev site and production when logged in as an admin user, while it no longer exists locally.
This PR removes the code that references the flag, but doesn't remove the flag itself from the dev and production environments yet. After this is merged and deployed, the flag can be safely deleted via the UI (shown in the photo) on the dev site and later on production.

Screenshot 2026-06-23 at 3 01 00 PM

SAUMILDHANKAR: Thanks for your guidance Whitney, sorry I am a little confused by the line "no longer exists locally." To make sure it doesn't work locally, I plan to visually confirm that after adding your code in my local, the buttons that are currently hidden (behind open now, more filters etc) are actually not visible (maybe by removing those buttons on top) but I am blanking on how to use featureflag name in .env.local.
Maybe I am overdoing it, plz ignore me lol

whitneywind: No worries, I appreciate your thoroughness! Could you please let me know which file(s) these currently hidden buttons are in?

SAUMILDHANKAR: Apologies for bearing with me, you are way more thorough. I am unable to find the hidden buttons referenced in the meeting. My understanding was that "Search this area" button could be hidden but that appears to have been removed long way back. The other button could be navigation compass but not sure :( 
So looks good removing legacy code, thanks.

client/src/components/FoodSeeker/SearchResults/ResultsFilters/ResultsFilters.tsx
```diff
@@ -52,45 +50,37 @@ const ResultsFilters: FC<ResultsFiltersProps> = ({
...
-                  isListView={showList}
-                  onClick={toggleShowList}
-                />
+              <Box sx={{ flex: 1, minWidth: 0 }}>
+                <AddressDropDown autoFocus={false} />
+              </Box>
+
+              <Box display={{ xs: "flex", sm: "none" }} alignItems="center" gap="0.5rem">
+                <Geolocate />
+                <SwitchViewsButton isListView={showList} onClick={toggleShowList} />
```
SAUMILDHANKAR, line 81: For now I can't think of a case where keeping Geolocate and SwitchViewsButton inside a box would be helpful.

SAUMILDHANKAR (approved):
Covers all the requirements of the issue. 👍

PR #2856, opened 2026-08-18
2117 add mvp full text search listings; "and" feature

### LINKED ISSUE #2117: Feature: Full text search for our listings
This would involve some dev search work, then some UI work to create the filter UI

Impact 4/5 (where 5 is very impactful) - this would help greatly for people looking for specific terms like 'senior' or 'veteran' or "First Lutheran Church of Inglewood" or "kosher"

### How is it now?

Currently, there is no way to search for a term in our database, other than with simple tag filters (pantry, meal), location, and hours.

There's no way to search for an organization name, nor what services it provides.

Also, much of our content is in free-form text descriptions ("We provide services only to seniors, veterans"), and not explicitly tagged ('baby diapers available'), so while searching for tags is nice, there is still a huge gap.
 
### Proposed new feature

Allow the user to search for text in the organization's listing's text fields
- Org name
- Address
- Phone number
- Requirements
- Email
- Notes
- Services
- Items available
- etc.

### Discussion

MVP is an "AND" search — all words entered must be present (non-case-specific).

Stretch: It would be amazing if we could also support some level boolean searches and quotes:

Most important, in my mind:
1. Quotes (for exact text (but not case-specific)), Exclude (i.e. -cheese)
2. OR (i.e. Parenthesis and commas between items)

We could use the Google search operators, since Google is so popular: i.e. "diapers" -cheese (kosher, -pork) 
https://southern.libguides.com/google/boolean

alternatively, or as MVP, a structured search that brea is out each field could also work.

# Specification

**MVP option: Text filter UI if we do full-text search (easy)**
image

- entering words into the text search will search ALL fields for that word, and filter results to only listings containing that word.   E.g. entering “army” will show “salvation **army**” locations as well as locations on “**army** street” that meet all other filter criteria. 

(Stretch option) if we can do Google-style Boolean search, please do that. E.g. “army -salvation” would filter to all locations that mention “army” but omit “Salvation” Army locations.

**Another option, if the above is too hard: Filter if we do structured search (harder because data is not tagged like this, but possibly more useful and specific):**

(same as above but at bottom… this instead of that single text field)

Name: (Boolean text search)
Address: (Boolean text search)
Phone number: (Boolean text search)
Eligibility: (Boolean text search)

Languages:
(Full checklist of tags used in this field for this region)

Notes: (Boolean text search)
Covid Notes: (Boolean text search)
Services:
(Full checklist of tags used in this field for this region)

Items available:
(Full checklist of tags used in this field for this region)

Something kind of like this but see above for full details:
image

**displaying search results:**
MVP: show matching results and put at top of search results "Searching for ‘X’" (where X is the search term the user entered, such as “Searching for ‘army -salvation’”)

If possible (this is probably very hard): highlight or bold matching text in the search results, including surfacing the matching text snippet to the top level search results when necessary.  

See Google for an example.  This would make it even more user-friendly.
image

VirginiaWu11:
- Closes #2117 

Summary
Add full-text search across public-facing listing fields
Update the filter UI from “Organization Name” to “Search”
Add Playwright coverage for address, multi-word, and cross-field searches

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.tsx
+++ b/client/src/components/FoodSeeker/SearchResults/ResultsFilters/FilterPanel.tsx
@@ -172,10 +172,10 @@ const FilterPanel: FC<FilterPanelProps> = ({ mealPantry, filterCount }) => {
       <Box sx={{ padding: "0 1rem 1rem 1rem", overflowY: "auto" }}>
         <Box sx={{ paddingBottom: "1rem" }}>
           <Typography variant="h4" sx={yPadding}>
-            Organization Name
+            Search
           </Typography>
           <OutlinedInput
-            placeholder="i.e. church of"
+            placeholder="i.e. kosher, senior, First Baptist, 90015"
             value={orgNameFilter}
             onChange={(e) =>
               dispatch({
@@ -186,7 +186,7 @@ const FilterPanel: FC<FilterPanelProps> = ({ mealPantry, filterCount }) => {
             endAdornment={
               <InputAdornment position="end">
                 <IconButton
-                  aria-label="Clear organization name filter"
+                  aria-label="Clear search filter"
                   edge="end"
                   onClick={() =>
                     dispatch({
--- a/client/src/hooks/useOrganizationBests.ts
+++ b/client/src/hooks/useOrganizationBests.ts
@@ -37,6 +37,24 @@ interface SelectAllParams {
   categoryIds: number[];
 }
 
+// Public-facing text fields searched by the "orgNameFilter" free-text search
+// (label reads "Search" in the UI). Deliberately excludes internal/admin-only
+// fields such as `adminNotes` -- only fields a food seeker can already see on
+// the listing/detail page belong here.
+const SEARCHABLE_FIELDS = [
+  "name",
+  "address1",
+  "address2",
+  "city",
+  "zip",
+  "phone",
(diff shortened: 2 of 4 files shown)


hanapotski (approved):
Approving — this is a clean, well-scoped, client-only change with no breaking impact. No server/API/DB/auth/SQL surfaces are touched, all searchable fields are type-safe against the `Stakeholder` type, the reducer wiring (`orgNameFilter`) is unchanged, and the 3 new Playwright tests line up correctly with the mock data (the `requirements` edit is on Stakeholder 2 / id 4408). Nice touch documenting the deliberate exclusion of `adminNotes` and other internal fields.

A few **non-blocking** notes for a future pass (none need to hold up merge):

1. **Naming vs. label drift** — the UI now reads "Search", but the underlying reducer state key and clear action are still `orgNameFilter` / "Clear organization name filter". Purely cosmetic, but renaming these later would keep the code matching the new UX.

2. **Search breadth is now much wider** — matching across `phone`, `email`, `notes`, `requirements`, `services`, `items`, etc. means a common word can now match more listings than the old name-only search. This is the intended feature (#2117), just worth being aware that result sets will be broader.

3. **Micro-perf** — the filter rebuilds an 11-field lowercased+joined string per stakeholder on every filter pass. Negligible at current dataset sizes (thousands of orgs); no change needed, just noting it if the list ever grows large or the filter runs on every keystroke.

---

Issue #10, opened 2019-10-16
Create Web API endpoints for resources & organizations


entrotech:
### Overview 
We need to setup the endpoints

### Resources
similar to what Fang did for the FAQ

ExperimentsInHonesty:
@lucas-homer eta tuesday!

PR #2859, opened 2026-08-22
fix(security): stop GET /accounts/:email leaking password hashes

hanapotski:
## Summary

Fixes **security audit finding #1 (Critical)** — the unauthenticated `GET /api/accounts/:email` endpoint returned the full account row, **including the bcrypt `password_hash` and all privilege/role flags**, to any caller. This enabled offline password cracking and account enumeration.

## Approach

`selectByEmail` is shared with internal callers (`authenticate`, profile update) that legitimately need the hash for `bcrypt.compare`, so it can't be stripped at the SQL layer. Instead the strip is done **at the service boundary** (per the audit's "not just by omission at the controller"):

- New `PublicAccount` type — only non-sensitive identity fields.
- New `accountService.selectByEmailPublic()` — returns `id, firstName, lastName, email, emailConfirmed, dateCreated` only.
- `accountController.getByEmail` now calls it, so the hash and privilege flags **can never reach a client** regardless of the caller.

## Compatibility

Both existing callers keep working:
- **Admin `Features.tsx`** — only uses `data.id`
- **Public `ForgotPassword.tsx`** — only uses the 200-vs-404 existence signal

## Tests

- Updated the `getByEmail` unit test to assert the sanitized shape and added `.not.toHaveProperty` guards for `passwordHash` / `isAdmin` / `isGlobalAdmin` to catch regressions.
- `tsc --noEmit` passes clean.
- Note: the `account.test.ts` suite has a **pre-existing** `Object {` vs `{` Jest-serializer snapshot mismatch (10 snapshots, identical on `develop`) unrelated to this change.

## Scope

Intentionally scoped to the **hash leak (#1)** only. Adding auth to this route would break the public forgot-password existence check; the unauthenticated-access / enumeration concern is a separate audit finding (**#10**) with its own recommended fix.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -262,3 +262,4 @@ node_modules/
 /client/playwright-report/
 /client/blob-report/
 /client/playwright/.cache/
+/.cursor
--- a/server/__test__/account.test.ts
+++ b/server/__test__/account.test.ts
@@ -252,34 +252,40 @@ describe("Account", () => {
       params: { email: "admin@test.com" },
     });
     const next = mockNext();
-    const selectByEmailMock =
-      accountService.selectByEmail as jest.MockedFunction<
-        typeof accountService.selectByEmail
+    const account = accounts[2];
+    // The controller must use the public-safe lookup, which never exposes the
+    // bcrypt password hash or privilege flags (security audit finding #1).
+    const selectByEmailPublicMock =
+      accountService.selectByEmailPublic as jest.MockedFunction<
+        typeof accountService.selectByEmailPublic
       >;
-    selectByEmailMock.mockImplementationOnce(() =>
-      Promise.resolve(accounts[2])
+    selectByEmailPublicMock.mockImplementationOnce(() =>
+      Promise.resolve({
+        id: account.id,
+        firstName: account.firstName,
+        lastName: account.lastName,
+        email: account.email,
+        emailConfirmed: account.emailConfirmed,
+        dateCreated: account.dateCreated,
+      })
     );
     await accountController.getByEmail(req, res, next);
-    expect(selectByEmailMock).toHaveBeenCalledWith("admin@test.com", 1);
-    expect(selectByEmailMock).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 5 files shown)


VirginiaWu11:
Reviewed this (fetched the branch into a worktree and independently verified rather than just reading the description):

**Fix is sound.** The unauthenticated `GET /api/accounts/:email` previously returned the full `login` row, including `password_hash` and every privilege flag. `selectByEmailPublic()` correctly strips this at the service layer (not just controller-side omission), reusing the existing `selectByEmail` SQL rather than duplicating it, and the controller now calls the sanitized version. No role/auth checks were touched, no new SQL string interpolation introduced.

**Verified locally on the branch:**
- `tsc --noEmit` — clean
- `npm run lint` (server) — clean
- `prettier --check` on all 4 changed server files — clean
- `npx jest --ci __test__/account.test.ts` — isolating `returns an account based on email`, the new `not.toHaveProperty("passwordHash"/"isAdmin"/"isGlobalAdmin")` guards all **pass**. The only failure is the inline-snapshot `Object {}` vs `{}` serializer mismatch, which is the pre-existing, already-documented Jest/pretty-format issue on `develop` — confirmed unrelated to this change, not something it introduced or needs to fix.
- Checked both client callers named in the description: `Features.tsx` only reads `data.id`, `ForgotPassword.tsx` only relies on the 200/404 existence signal — both are unaffected by the trimmed response shape.

**Scoping:** Deliberately not adding auth to the route (would break the public forgot-password existence check) and deferring the enumeration concern to a separate finding is the right call — flagged rather than silently bundled in, which matches how this should be handled.

**Nit:** the `.gitignore` hunk (adding `/.cursor`, missing trailing newline) is unrelated to the security fix and looks like it rode along from local tooling — harmless, but worth a note so it doesn't get mistaken for part of the security change.

Looks good to merge from a correctness/security standpoint.

VirginiaWu11 (approved):
Approving based on the review above — fix is correctly scoped, verified typecheck/lint/prettier clean, and the only failing test is the pre-existing unrelated snapshot-format issue on `develop`.

PR #2860, opened 2026-08-22
2117 rename orgNameFilter -> searchFilter to match "Search" UI label

VirginiaWu11:
Follow-up to #2856.

hanapotski's review on #2856 approved the PR but flagged that internal
state/action names (`orgNameFilter`, `ORG_NAME_FILTER_UPDATED`) still
referenced "org name" even though the field now searches multiple listing
fields and the UI label was changed to "Search".

## What changed
Mechanical rename across state/reducer/hook/action-type/call-sites:
- `orgNameFilter` -> `searchFilter`
- `ORG_NAME_FILTER_UPDATED` -> `SEARCH_FILTER_UPDATED`
- `useOrgNameFilter` -> `useSearchFilter`

The shareable `?name=` URL query parameter key is unchanged, so existing
shared/bookmarked search URLs keep working.

## Not addressed (both explicitly non-blocking in the original review)
- "Expanded matching scope" observation describes intended behavior from
  #2117, not something to change.
- Performance note on rebuilding the per-stakeholder searchable string on
  each filter pass is consistent with how every other filter in
  `useOrganizationBests.ts` already works, and the reviewer called it
  negligible at current scale / worth monitoring, not urgent.

## Checks
- Rebased onto latest `develop`, dependencies reinstalled from the
  committed lockfile (`npm ci` in `client/` and `server/`, no lockfile
  drift).
- `npm run typecheck` (client): **0 errors**, clean. (An earlier run against
  a stale local `node_modules` had shown 144 errors attributed to MUI
  v5→v7 migration debt — that was an artifact of a stale install, not a
  real issue; a clean `npm ci` install typechecks cleanly.)
- `npx playwright test --project=chromium tests/organizations.spec.ts`:
  11/11 passed, re-verified after the rebase + reinstall.

### DIFF
--- a/client/src/appReducer.tsx
+++ b/client/src/appReducer.tsx
@@ -145,10 +145,10 @@ function filterPanelReducer(state: boolean, action: AppAction): boolean {
       return state;
   }
 }
-function orgNameFilterReducer(state: string, action: AppAction): string {
+function searchFilterReducer(state: string, action: AppAction): string {
   switch (action.type) {
-    case "ORG_NAME_FILTER_UPDATED":
-      return action.orgNameFilter;
+    case "SEARCH_FILTER_UPDATED":
+      return action.searchFilter;
     default:
       return state;
   }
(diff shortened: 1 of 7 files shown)


hanapotski (approved):
Approving — this cleanly resolves the naming-drift note from the #2856 review.

Verified:
- **Complete rename** — grepped the entire `client/` tree (including tests) for `orgNameFilter` / `ORG_NAME_FILTER` / `useOrgNameFilter`: zero remaining references. That's the main risk for a rename touching a stringly-typed reducer (a missed dispatch site would silently break the filter with no compile error), so it's the key thing to get right, and it's fully covered.
- **Consistent wiring** — the `AppAction` union and `AppState` key are both updated, the reducer reads `action.searchFilter`, and both dispatch sites (`FilterPanel.tsx`, `SearchResults.tsx`) send the matching shape.
- **No behavior change / backward compatible** — the `?name=` URL query key is deliberately unchanged, so existing shared/bookmarked search URLs keep working.

Also agree with leaving the two deferred notes from the original review as-is: the expanded matching scope is the intended #2117 behavior, and the per-pass searchable-string rebuild is consistent with how the other filters in `useOrganizationBests.ts` already work and is negligible at current scale.

---

PR #2863, opened 2026-08-25
security fix in stakeholder details

whitneywind:
- Removed dangerouslySetInnerHTML from stakeholder notes/requirements fields to fix the stored XSS
- Rewrote the phone/email linkifier to render React elements instead of an HTML string
- Fixed a related bug where the old two-pass regex could mess up overlapping matches

### DIFF
--- a/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.tsx
+++ b/client/src/components/FoodSeeker/SearchResults/StakeholderDetails/StakeholderDetails.tsx
@@ -33,7 +33,7 @@ import ForkIcon from "icons/ForkIcon";
 import AppleIcon from "icons/AppleIcon";
 import IosShareIcon from "@mui/icons-material/IosShare";
 import SubdirectoryArrowRightIcon from "@mui/icons-material/SubdirectoryArrowRight";
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo, useState, type ReactNode } from "react";
 import { useNavigate } from "react-router-dom";
 import * as analytics from "services/analytics";
 import {
@@ -99,7 +99,6 @@ const StakeholderDetails = ({
   const navigate = useNavigate();
   const { setToast } = useToasterContext();
   const { tenantTimeZone } = useSiteContext();
-  const [paddingBottom, setPaddingBottom] = useState(30);
 
   if (!selectedOrganization) {
     return null;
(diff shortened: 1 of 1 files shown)


hanapotski (commented):
## Review

**The security fix itself is correct and well-done.** There's one blocking logistics problem (stale base) and a couple of minor nits.

The change removes `dangerouslySetInnerHTML` from the Notes / Closure Notes / Hour Notes / Eligibility fields and rewrites `formatEmailPhone` to return `ReactNode[]` instead of an interpolated HTML string.

### ✅ What's right

- **The XSS is genuinely fixed.** All four `dangerouslySetInnerHTML` usages are gone, and phone/email values are now rendered as React children, which React auto-escapes. Verified a `<script>alert(1)</script>` payload through the new logic — it comes out as a plain text node, not markup.
- **The linkifier rewrite is correct.** The old two-pass `text.replace()` could corrupt overlapping matches; the new single-pass `combinedRegex.exec()` loop with `lastIndex` tracking is the right pattern. Tested against mixed phone/email input, adjacent matches, no-match text, empty string, and `+1 (800) 555-1234` — all produce correct node arrays.
- `isEmail = match[1] !== undefined` correctly distinguishes the email vs. phone alternation group. Anchor keys are unique; plain-string nodes don't need keys.
- **Removing the `paddingBottom` state is safe** — it's unused anywhere on the branch.

### 🔴 Blocking: the branch is stale and won't merge cleanly

This branch is based on an older `develop`. Since then, this exact file was heavily refactored on `develop` — the `Grid2` component was removed entirely. The PR's copy still:

- imports `Grid2` from `@mui/material/Unstable_Grid2/Grid2`, which **no longer resolves** against current `develop`'s MUI (fails `tsc` with `TS2307`), and
- uses `Grid2` in 7 places that no longer exist on `develop`.

The PR's *own* changes typecheck fine — the only error is this stale `Grid2` import inherited from the old base. **Please merge/rebase current `develop`** and re-apply the security change onto the refactored file. Expect merge conflicts. As-is it won't build against `develop`.

### 🟡 Minor

- **Indentation glitch** around the new loop: the `while` closing brace and the trailing `if (lastIndex < text.length)` / `return nodes` are over-indented so they *look* nested inside the loop (they aren't — braces balance and it compiles). Worth running Prettier.
- `if (!text) return [text]` returns `[""]` for empty input — harmless; `[]` would be marginally cleaner.
- Pre-existing regex limitations (TLD capped at `\w{2,4}`, adjacent email-immediately-before-phone) are unchanged — out of scope, fine to leave.

### Recommendation

Approve the security approach, but please **merge/rebase `develop` and resolve the `Grid2` refactor conflict**, then re-run `tsc`. Once it typechecks against current `develop`, it's good to merge.

---
*This review was written by Claude on behalf of @hanapotski.*

hanapotski (approved):
## Approving ✅

Re-reviewed after the rebase. The one blocker from the previous review — the stale base carrying a `Grid2` import that no longer resolved against current `develop` — is resolved: the branch was rebased onto the current `develop` tip, `Grid2` is gone from the file, and merge state is clean.

Verified against current `develop`:
- **XSS fix intact** — no `dangerouslySetInnerHTML` remain; all four fields (Notes, Closure Notes, Hour Notes, Eligibility/Requirements) render via `formatEmailPhone` returning `ReactNode[]`, which React auto-escapes.
- **Typecheck clean** — applied the PR's file over current `develop` and ran `tsc --noEmit`: no errors (the previous `TS2307` on the stale import is gone).
- **Linkifier rewrite correct** — single-pass `combinedRegex.exec()` with `lastIndex` tracking; the old two-pass `text.replace()` overlap bug stays fixed.

Non-blocking nits (fine to leave): `if (!text) return [text]` returns `[""]` for empty input (`[]` would be marginally cleaner); pre-existing regex TLD cap is out of scope.

Good to merge.

---
*This review was written by Claude on behalf of @hanapotski.*

PR #2868, opened 2026-08-25
Require JWT secret configuration

junjun107:
# 7 on the list

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ services:
       POSTGRES_HOST: db
       POSTGRES_PORT: 5432
       WAIT_FOR: db:5432
+      JWT_SECRET: local-dev-secret
 
     links:
       - db
--- a/server/__test__/jwt-session.test.ts
+++ b/server/__test__/jwt-session.test.ts
@@ -1,0 +1,32 @@
+describe("JWT secret configuration", () => {
+  const originalJwtSecret = process.env.JWT_SECRET;
+
+  afterEach(() => {
+    if (originalJwtSecret === undefined) {
+      delete process.env.JWT_SECRET;
+    } else {
+      process.env.JWT_SECRET = originalJwtSecret;
+    }
+    jest.resetModules();
+  });
+
+  it("fails when JWT_SECRET is missing", () => {
+    delete process.env.JWT_SECRET;
+
+    expect(() => {
+      jest.isolateModules(() => {
+        require("../middleware/jwt-session");
+      });
+    }).toThrow("JWT_SECRET environment variable is required");
+  });
+
+  it("loads when JWT_SECRET is configured", () => {
+    process.env.JWT_SECRET = "test-secret";
+
+    expect(() => {
+      jest.isolateModules(() => {
+        require("../middleware/jwt-session");
+      });
+    }).not.toThrow();
+  });
+});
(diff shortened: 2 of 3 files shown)


hanapotski (commented):
**Verdict: Approve, with a couple of operational follow-ups to confirm before/at merge.**

This is a clean, correct security fix. Removing the hardcoded JWT signing-secret fallback and failing fast at startup when `JWT_SECRET` is unset is the right posture — a committed fallback secret in a public repo means anyone can forge valid JWTs for any role, so fail-closed beats silently signing with a weak key.

### What's good
- **Minimal and focused** — one meaningful line changed, no scope creep.
- **Fail-fast at startup** rather than lazily at first sign/verify. Since the secret is a hard requirement, crashing on boot is the correct failure mode.
- **Regression tests included**, covering both the missing and configured cases, using `jest.isolateModules` so the tests don't pollute each other's module state. Location/naming matches the existing suite.
- **No collateral test breakage**: only routers import `jwt-session`, and the existing test files import controllers directly (not routers), so the new throw won't break the existing suite at import time.

### Things to confirm before merging (not code defects — consequences of fail-closed)

The server now refuses to boot without `JWT_SECRET`, so every environment that runs it needs the var set. Two gaps:

1. **`docker-compose.yml` will crash-loop.** The `api` service sets `DATABASE_URL`, `POSTGRES_*`, and `WAIT_FOR` but **no `JWT_SECRET`**. After this change, `docker-compose up` (a documented local workflow) throws at startup. Consider adding `JWT_SECRET` to that service's `environment:` block in this PR so the local Docker path keeps working.

2. **Heroku config vars.** Each auto-deploy app (`foodoasis`, `foodoasisdev`, `foodoasisvite`) must already have a `JWT_SECRET` config var set, or the next deploy crash-loops. Worth verifying in Heroku config before merging.

3. **Local dev onboarding (minor).** Contributors running the server locally now need `JWT_SECRET` in `server/.env`. There's no `.env.example` to signal that — a one-line doc note or sample env entry would save newcomers a confusing "why won't the server start." Fine as a follow-up.

### Nitpick (optional)
Good that the weak fallback string is gone. Keep it out of the PR/commit text verbatim as this lands in the public repo — the current commit message is appropriately generic, so no change needed.

Net: the code change itself is correct and I'd approve it. The one gap inside the diff is that `docker-compose.yml` should also get a `JWT_SECRET` so the fail-fast doesn't break the documented local container workflow; the Heroku check is an out-of-band confirmation.

---
*Written by Claude on behalf of @hanapotski.*

hanapotski (approved):
**Approving.** The updated PR resolves the one substantive concern from the prior review.

The new commit adds `JWT_SECRET: local-dev-secret` to the `api` service in `docker-compose.yml`, so the documented `docker-compose up` local workflow keeps working now that the server fails fast when `JWT_SECRET` is unset. Placement is correct (on the `api` service, and an obviously-local throwaway value). The core fix and the regression tests are unchanged from the version reviewed earlier — removing the `"mark it zero"` hardcoded fallback and throwing at module load when the secret is missing.

One operational item to confirm outside this PR (not a code defect): each auto-deploy Heroku app (`foodoasis`, `foodoasisdev`, `foodoasisvite`) must already have a `JWT_SECRET` config var set, or the next deploy will crash-loop. A `.env.example` for local onboarding would also be a nice follow-up, but is out of scope here.

---
*Written by Claude on behalf of @hanapotski.*

PR #2870, opened 2026-08-26
fix(security): require auth on GET /accounts/:email (account enumeration)

hanapotski:
## Summary

`GET /api/accounts/:email` had no auth middleware, so it was reachable anonymously. A prior fix (audit finding #1) stopped it from leaking the bcrypt password hash, but the endpoint stayed public — so it could still be used to confirm whether a given email has an account (`200` vs `404`) and to read that account's `id`, name, confirmation status and creation date. This closes that gap (**security audit finding #10**).

## Changes

**Server** — [`server/app/routes/account-router.ts`](https://github.com/hackforla/food-oasis/blob/security-fix-account-email-enumeration-10/server/app/routes/account-router.ts)
- Gate the route with `jwtSession.validateUserHasRequiredRoles`, using the **same role set as `GET "/"`** (`admin`, `security_admin`, `data_entry`, `global_admin`). `GET "/"` already exposes strictly more account data to those roles, so this adds no new exposure for authenticated staff while removing all anonymous access.

**Client** — [`client/src/components/Account/ForgotPassword.tsx`](https://github.com/hackforla/food-oasis/blob/security-fix-account-email-enumeration-10/client/src/components/Account/ForgotPassword.tsx)
- Remove the per-keystroke existence check that was the only public caller. It relied on the endpoint being anonymous and is redundant — the "account not found, please register" feedback is already shown on submit via the `forgotPassword` response (`FORGOT_PASSWORD_ACCOUNT_NOT_FOUND`).
- The admin **Features** screen (`isAdmin`) is unaffected; it sends the `jwt` cookie automatically.

**Tests** — [`server/__test__/account.test.ts`](https://github.com/hackforla/food-oasis/blob/security-fix-account-email-enumeration-10/server/__test__/account.test.ts)
- Add tests asserting the route's role guard rejects an unauthenticated caller (401), rejects an authenticated caller without a permitted role (401), and lets an admin through.

## Verification
- Server `typecheck` ✅ · Server `lint` ✅ · Client `typecheck` ✅
- New finding-#10 tests: 3/3 pass. (The 10 pre-existing stale-snapshot failures in `account.test.ts` — `Object {` vs `{` pretty-format mismatch — are documented in `CLAUDE.md` §7 and unrelated to this change.)

## Note
A related but **separate** account-enumeration vector exists in the `POST /forgotPassword` flow (distinct response codes reveal account existence). It's intentionally out of scope here and tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2871, opened 2026-08-26
fix(security): stop forgotPassword revealing whether an email is registered

hanapotski:
## Summary

`POST /api/accounts/forgotPassword` returned a distinct `FORGOT_PASSWORD_ACCOUNT_NOT_FOUND` response for an unregistered email and a success response for a registered one, so an anonymous caller could use it to probe which emails have accounts (**account enumeration**). This is the follow-up flagged in #2870 — a separate vector from the `GET /accounts/:email` fix.

## Change (server-only)

[`server/app/services/account-service.ts`](https://github.com/hackforla/food-oasis/blob/security-fix-forgot-password-enumeration/server/app/services/account-service.ts) — `forgotPassword` now always returns the **same generic success response** whether or not the email is registered:
- A reset token is generated and the email sent **only when a matching account exists**.
- A genuine email-send failure (only reachable for a registered address) is **logged server-side** instead of returned, so the outcome stays indistinguishable from an unregistered email.

No client change is required: the success path already routes to the "reset link sent" confirmation screen, whose wording makes the same claim regardless of account existence.

## Trade-off
Users who mistype their email no longer get an explicit "no account — please register" message; they see the generic "if an account exists, a link was sent" confirmation. This is the standard anti-enumeration behavior and was the chosen approach.

## Tests
[`server/__test__/account-service.test.ts`](https://github.com/hackforla/food-oasis/blob/security-fix-forgot-password-enumeration/server/__test__/account-service.test.ts) (new) asserts:
- Unregistered address → generic success, **no token stored, no email sent**.
- Registered address → reset email sent, **identical** response body.
- The two responses are byte-for-byte equal (caller can't tell them apart).
- A send failure still returns generic success (logged server-side).

`typecheck` ✅ · `lint` ✅ · new tests 4/4 ✅.

## Notes
- Independent of #2870 (server-only; no shared files), so the two can merge in either order.
- The now-unreachable `FORGOT_PASSWORD_ACCOUNT_NOT_FOUND` branch in `ForgotPassword.tsx` is harmless dead code left for a later cleanup to avoid overlapping with #2870.

### DIFF
--- a/server/__test__/account-service.test.ts
+++ b/server/__test__/account-service.test.ts
@@ -1,0 +1,112 @@
+import db from "../app/services/db";
+import { sendResetPasswordConfirmation } from "../app/services/ses-service";
+import accountService from "../app/services/account-service";
+
+// forgotPassword must not reveal whether an email is registered (account
+// enumeration). These tests assert the response is identical for a registered
+// and an unregistered address, and that a reset email is only sent when a
+// matching account actually exists.
+jest.mock("../app/services/db", () => ({
+  __esModule: true,
+  default: {
+    oneOrNone: jest.fn(),
+    none: jest.fn(),
+  },
+}));
+
+jest.mock("../app/services/ses-service", () => ({
+  __esModule: true,
+  sendResetPasswordConfirmation: jest.fn(),
+  sendRegistrationConfirmation: jest.fn(),
+}));
+
+const oneOrNoneMock = db.oneOrNone as jest.Mock;
+const noneMock = db.none as jest.Mock;
+const sendResetMock = sendResetPasswordConfirmation as jest.Mock;
+
+const clientUrl = "https://example.test";
+const genericResponse = {
+  isSuccess: true,
+  code: "FORGOT_PASSWORD_SUCCESS",
+  message:
+    "If an account exists for that email, a password reset link has been sent.",
+};
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+describe("forgotPassword account enumeration", () => {
+  it("returns a generic success without sending email for an unregistered address", async () => {
+    oneOrNoneMock.mockResolvedValueOnce(null);
+
+    const result = await accountService.forgotPassword({
+      email: "nobody@test.com",
+      clientUrl,
+    });
+
+    expect(result).toEqual(genericResponse);
+    // No token stored, no email sent -- nothing observable that would confirm
+    // whether the account exists.
+    expect(noneMock).not.toHaveBeenCalled();
+    expect(sendResetMock).not.toHaveBeenCalled();
+  });
+
+  it("sends the reset email and returns the same generic success for a registered address", async () => {
+    oneOrNoneMock.mockResolvedValueOnce({ id: 42 });
+    noneMock.mockResolvedValueOnce(undefined);
+    sendResetMock.mockResolvedValueOnce(undefined);
+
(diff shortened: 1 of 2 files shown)


junjun107 (commented):
## Review

**What changed** (138+/26−, 2 files): `server/app/services/account-service.ts`'s `forgotPassword` now always returns the same `FORGOT_PASSWORD_SUCCESS` generic response body; only when a matching account actually exists does it generate a token and send the email, and any send failure is logged server-side instead of surfaced. New test file `server/__test__/account-service.test.ts` (4 tests) exercises the service directly with mocked `db`/`ses-service`.

**Verified independently, not just from the PR description:**
- Confirmed the old behavior on `develop`: distinct `FORGOT_PASSWORD_ACCOUNT_NOT_FOUND` vs `FORGOT_PASSWORD_SUCCESS` responses — a real enumeration oracle, and exactly the vector #2870 flagged as out-of-scope.
- The new code correctly routes through the existing `requestResetPasswordConfirmation` helper (unchanged) for the token-insert + email-send, which is why the test's `db.none`/`sendResetPasswordConfirmation` mocks line up with what actually executes.
- "No client change required" claim checks out: `newId` from the response is declared in the client's `ApiResponse` interface but never read anywhere, and `FORGOT_PASSWORD_ACCOUNT_NOT_FOUND` is referenced only in the branch the PR correctly calls now-unreachable dead code (left alone to avoid overlap with #2870, as stated).
- The pre-existing `account.test.ts` "forgot password" test fully mocks `accountService.forgotPassword`, so it doesn't exercise the real logic this PR changed — confirmed it's unaffected.

**Checks run** (isolated worktree off this branch):
- `npm run typecheck` — clean. `npm run lint` — clean. Matches the PR's claims.
- `npx jest --ci __test__/account-service.test.ts` — 4/4 pass, matches claim.
- Full `npx jest --ci` — 31 total, 10 failed / 21 passed. Same known `account.test.ts` stale-snapshot baseline, no new failures.

**Finding — Major: response-timing side channel reintroduces the enumeration oracle this PR sets out to close.**

`server/app/services/account-service.ts` (the `forgotPassword` function) — for an unregistered email, the function returns almost immediately after one `SELECT`. For a registered email, it `await`s a DB insert *and* a synchronous call to SES (`sendResetPasswordConfirmation`) before returning. The response bodies are now identical — the new tests correctly verify that — but the *time* to get that response is not, and SES round-trips are typically slow and variable enough (vs. a single indexed `SELECT`) to be a distinguishable, measurable signal. A response-timing side channel can defeat "the caller cannot tell them apart" just as effectively as the old status-code difference did, just with a higher bar to exploit. This is a well-documented gap in the "always return the same body" mitigation pattern — OWASP's forgot-password guidance calls out response-time equivalence alongside response content.

Suggestion: don't block the response on the email send for the registered-account path — fire-and-forget `requestResetPasswordConfirmation` (log failures async, same as now) so both paths return at comparable speed, or pad the unregistered-path response with an artificial delay roughly matching the registered-path's typical latency. Either looks like a small, contained change to the same function.

No other findings — no SQL/auth/schema changes, diff hygiene is clean, test coverage for the intended fix is solid, and the trade-off called out in the PR body (users no longer get "please register" copy) is a reasonable, standard trade-off for this mitigation.

---
*Posted from an independent local review — happy to take another pass once the timing side channel is addressed.*

junjun107 (approved):

PR #2880, opened 2026-09-03
test: server unit tests, working client Jest, and working client lint

hanapotski:
## Summary

**Server**
- Adds 11 Jest suites (153 tests): JWT session middleware, request validation schemas, stakeholder row mapping, CSV import helpers, SQL literal helpers, and the suggestion, category, tag, parent organization, announcements, and feature controllers. All follow the existing mocked-service pattern in `server/__test__/`.
- Fixes the 10 long-standing failures in `account.test.ts`. They were purely a pretty-format mismatch (`Object { ... }` on Jest 28 vs `{ ... }` in the inline snapshots). Setting `snapshotFormat: { printBasicPrototype: false }` in `jest.config.ts` resolves it without rewriting any snapshots.

**Client Jest**
- Makes the two Jest unit tests run again. The Jest config sends every file through `babel-jest`, but `babel.config.json` had no TypeScript preset, so any `.ts`/`.tsx` import failed to parse. Adds `@babel/preset-typescript` (7.x, matching `@babel/core`) and pins Jest `roots` to `src/` so it no longer picks up the Playwright specs in `tests/`.
- `App.test.js` now stubs axios, `window.scrollTo`, and the analytics `dataLayer`, so the render makes no real network requests and logs nothing after completion. It also asserts the app actually mounts and posts its visit event.
- Adds a `test:unit` script (`jest --ci`), since `npm test` in the client launches Playwright.

**Client lint**
- Replaces the dead CRA-era `.eslintrc.json` (extending `react-app`, with no `eslint` installed) with a flat `eslint.config.mjs` on ESLint 9: `@eslint/js`, `typescript-eslint`, `eslint-plugin-react`, the two classic `react-hooks` rules, and `eslint-config-prettier` last so Prettier keeps owning formatting. `npm run lint` is now `eslint .`.
- Rule choices mirror the server: `no-explicit-any` off, unused *arguments* ignored but unused variables and imports are errors, `react/no-unescaped-entities` off for prose in JSX. Playwright specs are linted with `rules-of-hooks` off because `test.extend` fixtures take a `use` callback the rule mistakes for React's `use`.
- Fixes everything the new config reported as an error, including three real hooks-rule violations:
  - `StakeholderDetails` called `useMemo` and two `useEffect`s after an early `return null`. Split into a null-checking wrapper and a content component whose hooks always run.
  - `OrganizationEdit` called `useMemo` inside a Formik render prop. Now computed directly, since the callback re-runs every render anyway.
  - `VerificationAdminGridMui` returned anonymous DataGrid cell components; they are now named.
  - Plus unused imports/types/variables, empty object types, a useless try/catch, a stray non-breaking space, `prefer-const` and boolean-cast autofixes, and a sparse-array typo in `jest.config.cjs`.
- Lint reports 0 errors and 10 `react-hooks/exhaustive-deps` warnings, left as warnings for follow-up.

**Docs**
- `CLAUDE.md` sections 3, 7, and 10 updated to describe the working setup for both workspaces.

## Notes

- The announcements and feature controllers have a "record not found" branch on delete that the services can never trigger, since both `remove` functions always resolve with a status object. The tests cover the real contract; the dead branch is left as is.
- Vite's React plugin does not read `babel.config.json`, so the added preset only affects Jest.
- Several client files show Prettier-only diffs because they had never been formatted. `client/package.json` still carries an inert `react-scripts` entry under `overrides`; left for a dependency-focused change.

## Test plan

Server, from `server/`:
- [x] `npx jest --ci` — 14 suites, 183 tests, 10 snapshots pass
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean

Client, from `client/`:
- [x] `npm run test:unit` — 2 suites, 7 tests pass, no console noise
- [x] `npm run lint` — 0 errors, 10 warnings
- [x] `npm run typecheck` — clean

Written by Claude on behalf of @hanapotski


SAUMILDHANKAR:
In server, I get

```
npm run typecheck

> foodoasis-web-api@1.0.102 typecheck
> tsc --noEmit

app/services/aws-service.ts:6:39 - error TS2307: Cannot find module '@aws-sdk/client-cognito-identity' or its corresponding type declarations.

6 import { CognitoIdentityClient } from "@aws-sdk/client-cognito-identity";
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in app/services/aws-service.ts:6
```

so i will investigate this further

PR #2885, opened 2026-09-15
fix(security): CSV uploads

Giolii:
Fix security issue # 14 in the Security Audit


hanapotski:
## Code review

Small, well-scoped diff fixing security-audit issue #14: adds a multer `fileFilter` (extension + MIME allowlist) and 5MB size limit on `/api/import/stakeholders-csv`, proper error responses instead of silent hangs/swallowed errors, and a client-side `accept` hint on the file input. The core plumbing checks out — the `handleUploadError` middleware placed after `upload.single("file")` correctly intercepts multer's `next(err)` calls, and `parseCsv`'s new `throw` correctly propagates through the `for await` loop to the controller's `try/catch`.

A few findings from review, most significant first:

1. **CSV MIME-type allowlist can reject legitimate files** (`server/app/routes/import-router.ts:70`) — `CSV_MIME_TYPES` is a fixed set (`text/csv`, `application/vnd.ms-excel`, `text/plain`). Browsers/OSes commonly report `.csv` files as `application/octet-stream` or an empty string when the extension isn't MIME-registered locally, so a real admin's valid CSV that worked before this PR could now be rejected with "Only CSV files are allowed."

2. **CSV-content injection not addressed** (`server/app/routes/import-router.ts:58`) — the project's contributor guidance calls out validating "injection via CSV content" for changes to `import-controller.ts`. This PR only gates on filename extension + client-declared MIME type (both attacker-controlled/spoofable), not actual cell content, so formula-injection risk (e.g. a cell starting with `=`) remains open even though the PR's purpose is closing out this exact audit item.

3. **Fragile error routing** (`server/app/routes/import-router.ts:94`) — `handleUploadError` distinguishes the fileFilter's custom error from any other error via an exact string match on `err.message === "Only CSV files are allowed."`. If that literal string ever drifts between the two spots (typo, wording tweak, i18n), the branch silently falls through to `next(err)` instead of returning the intended 400.

4. **No new tests** for the added size/type validation or the new error-handling paths, despite this being a security-fix PR.

None of these are blockers on the happy path, but worth a look before merge — especially #1, which could break real uploads for some users.

---

Giolii:
Thanks for the review! I updated the MIME handling, replaced the string-based error check, and added tests. I propose tracking CSV formula injection in a separate issue since it’s outside the scope of this PR

hanapotski:
## Re-review (after `fix(security): bound CSV uploads` / `test(security): cover CSV upload validation`)

Checked out the updated branch in an isolated worktree and ran the new test suite: 8/8 passing (`npx jest __test__/import-controller.test.ts __test__/import-router.test.ts`).

Three of the four points from the earlier review are resolved:

- ✅ **MIME-type false rejection** — `application/octet-stream` was added to `CSV_MIME_TYPES`, with a new test covering the "browser doesn't report a MIME type" case that would previously have rejected a valid CSV.
- ✅ **Fragile error routing** — the exact-string match on `err.message` was replaced with a dedicated `InvalidCsvFileError` class and an `instanceof` check, so wording changes can't silently break the 400 response.
- ✅ **Test coverage** — `server/__test__/import-controller.test.ts` and `server/__test__/import-router.test.ts` now cover missing-file, parse-failure, oversized-file (413), and extension/MIME-mismatch paths.

Still open (non-blocking, worth a follow-up):

- **CSV-content injection** — the project's guidance for this file calls out validating injection via CSV *content*, not just file type. This update still only gates on filename extension + declared MIME type; cell values aren't sanitized before being stored/re-exported. Not something this PR needs to solve, but flagging so it doesn't get lost.

Looks good to merge on the file-upload hardening itself.

---

